PR #1, opened 2022-03-04
DM-33918: Create code repository for prompt processing

kfindeisen:
This PR adds in the code removed from lsst-dm/dmtn-219#2. The code has been cleaned up to pass `flake8`, but no attempt has been made to further reconcile it with LSST conventions such as namespacing.

### DIFF
--- a/src/tester/upload.py
+++ b/src/tester/upload.py
@@ -81,7 +81,11 @@ def main():
     )
     storage_client = storage.Client(PROJECT_ID, credentials=credentials)
     bucket = storage_client.bucket("rubin-prompt-proto-main")
-    publisher = pubsub_v1.PublisherClient(credentials=credentials)
+    batch_settings = pubsub_v1.types.BatchSettings(
+        max_messages=INSTRUMENTS[instrument].n_detectors,
+    )
+    publisher = pubsub_v1.PublisherClient(credentials=credentials,
+                                          batch_settings=batch_settings)
 
     blobs = storage_client.list_blobs(
         "rubin-prompt-proto-main",
--- a/src/tester/upload.py
+++ b/src/tester/upload.py
@@ -1,0 +1,110 @@
+from dataclasses import dataclass
+from google.cloud import pubsub_v1, storage
+from google.oauth2 import service_account
+import json
+import logging
+import random
+import sys
+import time
+from visit import Visit
+
+
+@dataclass
+class Instrument:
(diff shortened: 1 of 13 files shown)


src/tester/upload.py
```diff
@@ -83,9 +81,6 @@ def main():
     )
     storage_client = storage.Client(PROJECT_ID, credentials=credentials)
     bucket = storage_client.bucket("rubin-prompt-proto-main")
-    batch_settings = pubsub_v1.types.BatchSettings(
```
ktlim: We should probably leave this in and add it to the `PublisherClient` constructor below.

ups/prompt-prototype.table
```diff
@@ -4,5 +4,6 @@
 #   the "base" package.
 setupRequired(base)
 setupRequired(sconsUtils)
+setupRequired(daf_butler)    # Used by middleware_interface module.
```
ktlim, line 7: I don't think this is ever going to be an eups package, so I'm not sure the `ups` directory is needed at all.

ktlim (approved):
Thanks for cleaning up some of my leftovers.

src/tester/upload.py
```diff
@@ -83,9 +81,6 @@ def main():
...
-        max_messages=INSTRUMENTS[instrument].n_detectors,
-    )
     publisher = pubsub_v1.PublisherClient(credentials=credentials)
```
kfindeisen, line 84: @ktlim, is this what it should look like? Google's documentation is shockingly bad.
```suggestion
    publisher = pubsub_v1.PublisherClient(credentials=credentials, per_partition_batching_settings=batch_settings)
```

ktlim: See https://googleapis.dev/python/pubsub/latest/publisher/index.html#batching
and https://googleapis.dev/python/pubsub/latest/publisher/api/client.html#module-google.cloud.pubsub_v1.publisher.client

I think this is just `batch_settings=batch_settings`.

kfindeisen: Oops, wrong namespace: I was looking at https://cloud.google.com/python/docs/reference/pubsublite/latest/google.cloud.pubsublite.cloudpubsub.publisher_client.PublisherClient.

ktlim: https://cloud.google.com/pubsub/docs/choosing-pubsub-or-lite

PR #2, opened 2022-03-04
DM-33938: Rename prompt-prototype

kfindeisen:
This PR increases the degree to which this package conforms to the DM package template; the only elements still missing are an `lsst` namespace and a `doc/` directory.


parejkoj (commented):
playbook.rst should probably get moved to `doc/`?

Do we need bin.src or bin at all?

kfindeisen:
> playbook.rst should probably get moved to `doc/`?

I agree it shouldn't be in `python`, but I'm less sure about `doc`. I guess as long as we're not expecting to set up the usual Sphinx infrastructure?

> Do we need bin.src or bin at all?

Probably not, but I feel better having it ready in case something comes up. We are playing things by ear, after all.

parejkoj (approved):

PR #5, opened 2022-03-08
add Cloud Shell instructions

ebellm:


doc/playbook.rst
```diff
@@ -127,6 +127,15 @@ It accepts messages and gateways them to Cloud Run.
 tester
 ======
 
+Run the tester either on a local machine, or in Cloud Shell.
```
kfindeisen, line 130: I think the text defining what the tester _is_ really should go first; it's not a very intuitive name. I suggest moving the new text between "It takes a JSON token..." and "Command line arguments are...".

doc/playbook.rst
```diff
@@ -127,6 +127,15 @@ It accepts messages and gateways them to Cloud Run.
 tester
 ======
 
+Run the tester either on a local machine, or in Cloud Shell.  
+For the latter, install the protype code and the GCP pubsub client:
```
kfindeisen, line 131: Did you mean "for the former"? It seems strange that you'd have to do these steps (especially installing `google-cloud-pubsub`) on Google's own system, whereas they make sense on a local machine.

kfindeisen (commented):
Thanks for adding this, but can you please clarify whether the example you give is for a local machine or Cloud Shell?

doc/playbook.rst
kfindeisen: Never mind, these instructions are indeed for the Shell.

kfindeisen (approved):

kfindeisen:
Can you also add some information on how to acquire the JSON token in the Cloud Shell?

doc/playbook.rst
```diff
@@ -127,6 +127,15 @@ It accepts messages and gateways them to Cloud Run.
...
+
+.. code-block:: sh
+
+    gcloud config set project prompt-proto
+    git clone https://github.com/lsst-dm/prompt_prototype.git
+    pip3 install google-cloud-pubsub
+
 ``tester/upload.py`` is a script that simulates the CCS image writer.
 It takes a JSON token in ``./prompt-proto-upload.json``.
```
ktlim, line 140: We should modify it to not force the use of a JSON token when it's not necessary (such as in the Cloud Shell environment).

doc/playbook.rst
ktlim, line 131:
```suggestion
For the latter, install the prototype code and the GCP pubsub client:
```

PR #6, opened 2022-03-09
DM-33935: Create basic test framework for prompt-prototype

parejkoj:
This ended up being rather trivial; I'm putting this in review while I keep working on the flask/google.cloud tests (trying to mock a non-installed package (gcloud) is tricky). I'd rather have something to run in a github action now, while we figure out what we want the tests to actually do (I can't mock the butler interface until I have a better sense of how we'll actually use it).

### DIFF
--- a/python/activator/activator.py
+++ b/python/activator/activator.py
@@ -1,3 +1,26 @@
+# This file is part of prompt_prototype.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+__all__ = ["check_for_snap", "next_visit_handler"]
+
 import base64
 import json
 import logging
@@ -9,8 +32,8 @@ from typing import Optional, Tuple
 from flask import Flask, request
 from google.cloud import pubsub_v1, storage
 
-from middleware_interface import MiddlewareInterface
-from visit import Visit
+from .middleware_interface import MiddlewareInterface
+from .visit import Visit
 
 PROJECT_ID = "prompt-proto"
 
--- a/python/activator/middleware_interface.py
+++ b/python/activator/middleware_interface.py
@@ -68,7 +68,7 @@ class MiddlewareInterface:
         self.image_bucket = image_bucket
         self.instrument = instrument
 
-        self.repo = "/tmp/butler-{os.getpid()}"
+        self.repo = f"/tmp/butler-{os.getpid()}"
         if not os.path.exists(self.repo):
             _log.info(f"Making local Butler {self.repo}")
             Butler.makeRepo(self.repo)
--- a/python/activator/middleware_interface.py
+++ b/python/activator/middleware_interface.py
@@ -1,3 +1,26 @@
+# This file is part of prompt_prototype.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+__all__ = ["MiddlewareInterface"]
+
 import logging
 import os
 import shutil
@@ -5,7 +28,7 @@ import shutil
 from astropy.time import Time
 from lsst.daf.butler import Butler
 
-from visit import Visit
+from .visit import Visit
 
 PIPELINE_MAP = dict(
     BIAS="bias.yaml",
--- a/python/activator/middleware_interface.py
+++ b/python/activator/middleware_interface.py
@@ -41,6 +41,26 @@ _log.setLevel(logging.DEBUG)
 
 
 class MiddlewareInterface:
+    """Interface layer between the Butler middleware and the prompt processing
+    data handling system.
+
+    An instance of this class will accept an incoming group of snaps to
+    process, using an instance-local butler repo. The instance can pre-load
+    the necessary calibrations to process an incoming visit, ingest the data
+    when it is available, and run the difference imaging pipeline, all in that
+    local butler.
+
+    Parameters
+    ----------
+    input_repo : `str`
+        Path to a butler repo containing the calibration and other data needed
+        for processing images as they are received.
+    image_bucket : `str`
+        Google storage bucket where images will be written to as they arrive.
+    instrument : `str`
+        Name of the instrument taking the data, for populating butler
+        collections and dataIds.
+    """
     def __init__(self, input_repo: str, image_bucket: str, instrument: str):
 
         # self.src = Butler(input_repo, writeable=False)
@@ -106,6 +126,13 @@ class MiddlewareInterface:
         self.calib_types = ["bias", "dark", "defects", "flat", "fringe", ]
 
     def prep_butler(self, visit: Visit) -> None:
+        """Prepare a temporary butler repo for processing the incoming data.
+
+        Parameters
+        ----------
+        visit : Visit
+            Group of snaps from one detector to prepare the butler for.
+        """
         _log.info(f"Preparing Butler for visit '{visit}'")
         visit_info = visit.__dict__
         for calib_type in self.calib_types:
@@ -145,6 +172,14 @@ class MiddlewareInterface:
         shutil.rmtree(visit_dir, ignore_errors=True)
 
     def ingest_image(self, oid: str) -> None:
+        """Ingest an image into the temporary butler.
+
+        Parameters
+        ----------
+        oid : `str`
+            Google storage identifier for incoming image, relative to the
+            image bucket.
+        """
         _log.info(f"Ingesting image '{oid}'")
         run = f"{self.instrument}/raw/all"
         cmd = [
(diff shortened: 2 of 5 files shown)


python/activator/activator.py
```diff
@@ -1,3 +1,25 @@
...
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+__all__ = ["check_for_snap", "next_visit_handler"]
```
kfindeisen, line 21: Please include blank lines between the license blocks and the `__all__` definition; it's easy to miss the `__all__` with the current formatting.

python/activator/visit.py
```diff
@@ -1,3 +1,5 @@
+__all__ = ["Visit"]
```
kfindeisen, line 1: No license statement?

python/activator/activator.py
kfindeisen: Should `activator.py` export any components at all? It was pretty clearly written as a Python script, not a library (see the `if __name__ == "__main__":` block at the bottom).

python/activator/middleware_interface.py
```diff
@@ -40,6 +40,20 @@
 
 
 class MiddlewareInterface:
+    """Interface layer between the Butler middleware and the prompt processing
+    data handling system.
+
+    Parameters
```
kfindeisen, line 46: Can you include an explanation of what a `MiddlewareInterface` object actually represents? I find it very confusing that `__init__` performs Butler operations, some of them ones that change state.

The fact that some (but not all) of the code assumes multiple concurrent `MiddlewareInterface` objects is probably a clue...

python/activator/middleware_interface.py
```diff
@@ -40,6 +40,20 @@
...
+    ----------
+    input_repo : `str`
+        Path to a butler repo containing the calibration and other data needed
+        for processing images as they are received.
+    image_bucket : `str`
+        Datastore where images will be written to as the arrive.
```
kfindeisen, line 52:
```suggestion
        Datastore where images will be written to as they arrive.
```

kfindeisen: I'd also avoid saying "Datastore", since that's a Middleware term.

python/activator/middleware_interface.py
```diff
@@ -105,6 +119,13 @@ def __init__(self, input_repo: str, image_bucket: str, instrument: str):
         self.calib_types = ["bias", "dark", "defects", "flat", "fringe", ]
 
     def prep_butler(self, visit: Visit) -> None:
+        """Prepare a temporary butler repo to process the incoming data in.
```
kfindeisen, line 122:
```suggestion
        """Prepare a temporary butler repo for processing the incoming data.
```

python/activator/middleware_interface.py
```diff
@@ -105,6 +119,13 @@ def __init__(self, input_repo: str, image_bucket: str, instrument: str):
...
+
+        Parameters
+        ----------
+        visit : Visit
+            The visit to prepare the butler for.
```
kfindeisen, line 127: Note that a `Visit` object does not represent a visit, but (in Stack terms) an exposure. It is detector-specific, but not snap-specific.

python/activator/middleware_interface.py
```diff
@@ -144,6 +165,13 @@ def prep_butler(self, visit: Visit) -> None:
         shutil.rmtree(visit_dir, ignore_errors=True)
 
     def ingest_image(self, oid: str) -> None:
+        """Ingest an image into the temporary butler.
+
+        Parameters
+        ----------
+        oid : `str`
+            Identifier for incoming image.
```
kfindeisen, line 173: Might be worth documenting the format for the `oid` -- this seems to be the full bucket path to the image.

python/activator/middleware_interface.py
```diff
@@ -160,6 +188,15 @@ def ingest_image(self, oid: str) -> None:
         # subprocess.run(cmd, check=True)
 
     def run_pipeline(self, visit: Visit, snaps) -> None:
+        """Process the received image.
+
+        Parameters
+        ----------
+        visit : Visit
+            The visit information of the image to be processed.
```
kfindeisen, line 196: Again, a `Visit` is not a visit.

kfindeisen: Missing docstring for `filter_calibs`, below.

tests/test_middleware_interface.py
```diff
@@ -0,0 +1,72 @@
...
+
+import os.path
+import tempfile
+import unittest
+
+from activator.middleware_interface import MiddlewareInterface
+from activator.visit import Visit
+
+
+class MiddlewareInterfaceTest(unittest.TestCase):
```
kfindeisen, line 30: FWIW, I would prefer to keep using `unittest` for the test code just so that I don't have to learn a new system in a hurry. I agree that we don't need to support running tests as standalone scripts.

tests/test_middleware_interface.py
```diff
@@ -0,0 +1,72 @@
...
+    def setUp(self):
+        input_dir = os.path.join(os.path.abspath(os.path.dirname(__file__)), "data")
+        instrument = "TestCam"
+        self.output_dir = tempfile.TemporaryDirectory()
```
kfindeisen, line 34: Please include cleanup of this directory, either through `tearDown` or (my preference) `addCleanup`.

tests/test_middleware_interface.py
```diff
@@ -0,0 +1,72 @@
...
+        self.interface = MiddlewareInterface(input_dir, self.output_dir, instrument)
+        self.next_visit = Visit(instrument,
+                                detector=1,
+                                group=1,
+                                snaps=1,
+                                filter="r",
+                                ra=10,
+                                dec=20,
+                                kind="BIAS")
+        self.log_prefix = "INFO:activator.middleware_interface:"
```
kfindeisen, line 44: I'd suggest making an attribute be `activator.middleware_interface`, so that you don't have to hardcode it in `assertLogs` calls. Maybe `tested_logger_name`?

tests/test_middleware_interface.py
```diff
@@ -0,0 +1,72 @@
...
+        # Ideas for things to test:
+        # * On init, does the right kind of butler get created, with the right
+        #   collections, etc?
+        # * On init, is the local butler repo purely in memory?
+
+    def test_prep_butler(self):
+        with self.assertLogs("activator.middleware_interface", level="INFO") as cm:
+            self.interface.prep_butler(self.next_visit)
+        msg = f"{self.log_prefix}Preparing Butler for visit '{self.next_visit}'"
+        self.assertEqual(cm.output, [msg])
```
kfindeisen, line 60: `assertIn`? I think my changes on [DM-33931](https://jira.lsstcorp.org/browse/DM-33931) may make it hard to keep up this test style.

kfindeisen (approved):
Looks good; some style comments.

python/activator/visit.py
parejkoj: I didn't want to muck with the visit.py files, because we may want to consolidate those somehow (there's two copies in different places).

python/activator/middleware_interface.py
parejkoj: I don't think that's true? It has a `snaps` member, so it's a visit, not an exposure (assuming I'm understanding this correctly).

python/activator/activator.py
kfindeisen: I'm not sure I follow. You can import anything in Python, whether or not it's part of the module's API, and testing is a case where it's reasonable to import internals.

tests/test_middleware_interface.py
parejkoj: If we want to use pytest fixtures to test the flask part, I think we'll have to move at least partly away from unittest. Not sure how much yet.

tests/test_middleware_interface.py
parejkoj: `TemporaryDirectory` is automatically removed when the object is destroyed.

tests/test_middleware_interface.py
parejkoj: I hope to get rid of all of these assertLogs tests when I add butler API calls to do the work, so this is a short stopgap.

python/activator/visit.py
kfindeisen: This is the real one, the other is a symlink. I'm pretty sure we can delete the latter if we change the import in `upload.py`, I guess I was waiting for test coverage of `upload` before trying. 🙂

python/activator/activator.py
parejkoj: There are so many module-level items in `activator.py`, I wanted to hide them all.

python/activator/middleware_interface.py
kfindeisen: There is a separate `Visit` object for each detector; see https://github.com/lsst-dm/prompt_prototype/blob/main/python/tester/upload.py#L65 for the constructor call.

python/activator/middleware_interface.py
parejkoj: I've added a paragraph, but I think we don't want to assume that any of the stuff in `__init__` will stick around (I suspect most of that should move to `prep_butler`, but don't know yet).

python/activator/middleware_interface.py
parejkoj: Oh, right! It's a "detector-visit". It's not an "exposure" in stack terms (or even an `afw.image.Exposure`), because that's a single snap, or a combined snap, whereas this is the snap-group. Which, come to think of it, is I think called a group? I should ask about this: a `Visit` is the combed object, but we're getting all the snaps here.

Oof, our terminology here is so confusing.

python/activator/middleware_interface.py
parejkoj: I'm only going to document the things as we start to use them: this only appears in commented-out code.

python/activator/visit.py
parejkoj: Fair point, but I'm nervous about writing tests of `upload` (see KT's comment on DM-33938: `upload.py` is a test itself)

python/activator/middleware_interface.py
parejkoj: Lets defer on any changes to `visit` for now. I think we might eventually want to name this a `DetectorVisit`, but see the new thred in #dm-naming-things: we don't really have a good term for this.

python/activator/visit.py
kfindeisen: I disagree -- for now, at least, `upload.py` is the server for which `activator.py` is the client; that makes it part of the system we're developing. In particular, we probably will need to make changes to `upload.py` as we start doing more substantial integration testing.

python/activator/middleware_interface.py
kfindeisen: Of course, and we shouldn't document implementations anyway. I just meant that the class seems to be intended to be used in particular ways, which are not at all obvious from the name.

python/activator/middleware_interface.py
kfindeisen: I wasn't asking for changes to the `Visit` class, just that the documentation reflects that these are not full focal plane (which is what I assume when I hear "visit").

parejkoj:
Please have another look at the middleware docstrings: I've tweaked them per your suggestions.

kfindeisen (approved):
Looks good! Thanks for being patient with my pickiness. 🙂

python/activator/middleware_interface.py
```diff
@@ -137,7 +194,19 @@ def ingest_image(self, oid: str) -> None:
...
+
+        Parameters
+        ----------
+        visit : Visit
+            Group of snaps from one detector to be processed.
+        snaps : `set`
+            Identifiers of the snaps that were received.
+            TODO: I believe this is unnecessary because it should be encoded
+            in the `visit` object, but we'll have to test how that works once
+            we implemented this with actual data.
```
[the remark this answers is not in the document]
parejkoj: I feel like putting the TODO in the docstring makes us more likely to fix the docstring if we do remove it?

PR #9, opened 2022-03-10
DM-33931: Improve logging

kfindeisen:
This PR cleans up the logging to make it more user-friendly.

### DIFF
--- a/python/activator/activator.py
+++ b/python/activator/activator.py
@@ -98,12 +98,12 @@ def next_visit_handler() -> Tuple[str, int]:
         envelope = request.get_json()
         if not envelope:
             msg = "no Pub/Sub message received"
-            logging.warn(f"error: '{msg}'")
+            _log.warn(f"error: '{msg}'")
             return f"Bad Request: {msg}", 400
 
         if not isinstance(envelope, dict) or "message" not in envelope:
             msg = "invalid Pub/Sub message format"
-            logging.warn(f"error: '{msg}'")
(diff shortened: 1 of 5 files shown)


python/activator/activator.py
```diff
@@ -74,13 +74,13 @@ def check_for_snap(
     instrument: str, group: int, snap: int, detector: int
 ) -> Optional[str]:
     prefix = f"{instrument}/{detector}/{group}/{snap}/{instrument}-{group}-{snap}-"
```
timj, line 76: Do you want to make use of the MDC feature that we use in butler logging? That lets you set an MDC record and after that every log message will get that content included without you having to explicitly add "prefix" to every log message.

python/activator/middleware_interface.py
```diff
@@ -36,7 +36,7 @@
     FLAT="flat.yaml",
 )
 
-_log = logging.getLogger(__name__)
+_log = logging.getLogger("lsst." + __name__)
```
parejkoj, line 39: This uses `_log` here, but `logger` in `activator.py`; we should probably make those consistent, but I don't have a strong preference for one or the other (maybe `logger` because it doesn't require typing a `_`?).

timj: The underscore at least makes it clear that it's a private package variable.

kfindeisen: I'll change it to `_log`.

PR #10, opened 2022-03-10
DM-34018: Move activator service one level down.

ktlim:
This should enable relative paths to work properly.


kfindeisen (approved):
No questions about the changes, and no errors in 28 minutes of running. Thanks for fixing this!

PR #12, opened 2022-03-11
DM-34031: Add script for uploading raws.

kfindeisen:
This PR creates a script that can reproducibly upload raws from `ap_verify_ci_cosmos_pdr2`. It assumes that `gsutil` is already installed and configured.


bin/prompt_proto_upload_raws.sh
```diff
@@ -0,0 +1,33 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+# This script uploads the raw files from the ap_verify_ci_cosmos_pdr2 repo
+# to Google Storage. It renames the files to match prompt_prototype conventions.
```
parejkoj, line 24: Should note that the user has to have `setup ap_verify_ci_cosmos_pdr2` before this will work. I see there's an error message if it's not configured, but a note here would be good too.

bin/prompt_proto_upload_raws.sh
```diff
@@ -0,0 +1,33 @@
...
+
+set -e  # Abort on any error
```
parejkoj, line 26: Good practice!

bin/prompt_proto_upload_raws.sh
```diff
@@ -0,0 +1,33 @@
...
+
+RAW_DIR="${AP_VERIFY_CI_COSMOS_PDR2_DIR:?'dataset is not set up'}/raw"
+
+gsutil cp "${RAW_DIR}/HSC-0059150-050.fits.gz" \
+    gs://rubin-prompt-proto-unobserved/HSC/50/2016030700001/0/HSC-2016030700001-0-0059150-HSC-G-050.fits.gz
```
parejkoj, line 31: Maybe a comment above these that notes that the directory structure is built to mimic that of the butler.

parejkoj (commented):
Can we call it `prompt_prototype_upload_raws.sh`? I'd like to move away from "proto" if we can.

Or should we call it `prompt_prototype_upload_ci_cosmos_pdr2.sh`, to make it very explicit where it's coming from?

bin/prompt_proto_upload_raws.sh
kfindeisen: It doesn't mimic the Butler (in fact, I think K-T mentioned that it's deliberately different from the Butler). It mimics what's in `upload.py`/`activator.py`.

kfindeisen:
@parejkoj is this better?

PR #13, opened 2022-03-14
DM-33937: Setup a butler with real data on google cloud

kfindeisen:
This PR provides scripts for distilling `ap_verify_ci_hits2015` into only the data we want to put into the Prompt Processing "initial" repo. Note that `make_remote_butler.py` can only be run on Google Cloud.

### DIFF
--- a/bin.src/make_remote_butler.py
+++ b/bin.src/make_remote_butler.py
@@ -36,9 +36,9 @@ import argparse
 import logging
 import os
 import sys
-import time
 
 from lsst.utils import getPackageDir
+from lsst.utils.timer import time_this
 from lsst.daf.butler import Butler, CollectionType, Config
 from lsst.obs.base import Instrument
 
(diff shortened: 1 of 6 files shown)


bin.src/make_remote_butler.py
```diff
@@ -0,0 +1,77 @@
...
+both the target URI and the registry database.
+"""
+
+
+import argparse
+import logging
+import os
+import sys
+
+import lsst.log
```
timj, line 40: No butler code should be issuing lsst.log messages so please do not add lsst.log dependencies in code like this. In fact no python code should be using lsst.log and if there is something that is importing some code that sets the lsst.log logger we can talk about the best way to do it in lsst.utils.logging or using the butler code without making lsst.log visible anywhere.

bin.src/make_remote_butler.py
```diff
@@ -0,0 +1,77 @@
...
+    return parser
+
+
+def main():
+    logging.basicConfig(level=logging.INFO, stream=sys.stdout)
+    lsst.log.configure_pylog_MDC("DEBUG", MDC_class=None)
+
+    args = _make_parser().parse_args()
+    seed_config = lsst.daf.butler.Config(args.seed_config)
+    logging.info("Creating repository at %s...", args.target_repo)
```
timj, line 67: Why does the make_local_butler.py file exist? Isn't that identical to this version if the default value for `--target-repo` was "."?  Have you considered making this command a butler pluggable command so that you could make use of the normal `butler --help` click infrastructure? (it's butler create + butler import).

bin.src/make_remote_butler.py
```diff
@@ -0,0 +1,77 @@
...
+def _make_parser():
+    parser = argparse.ArgumentParser()
+    # Could reasonably be positional arguments, but keep them as keywords to
+    # prevent users from confusing --src-repo with --target-repo.
+    parser.add_argument("--src-repo", required=True,
+                        help="The location of the repository whose files are to be copied.")
+    parser.add_argument("--target-repo", required=True,
+                        help="The URI of the repository to create.")
+    parser.add_argument("--seed-config",
+                        default=os.path.join(os.environ["PROMPT_PROTOTYPE_DIR"], "etc", "db_butler.yaml"),
```
ktlim, line 54: https://pipelines.lsst.io/v/daily/py-api/lsst.utils.getPackageDir.html#lsst.utils.getPackageDir

ktlim (approved):
I agree with Tim's comments as well.

bin.src/make_remote_butler.py
kfindeisen: This is simply vestigial code; there's no reason to get so worked up about it.

bin.src/make_remote_butler.py
kfindeisen: I do not believe that integrating what should be a local, highly specific script into the `butler` ecosystem would be beneficial, even if we had the time to do so.

We are deliberately not using the `butler` command-line itself because we need to understand what our overheads are, and the `butler` utility loses a lot of time on package import and `Butler` setup.

PR #14, opened 2022-03-17
DM-34068: Implement Prompt Prototype's data transfer to workers

parejkoj:
This has a lot of TODOs and is not what we want, but I think it mostly works?

I'm still working on adding tests to it, but it should be reviewable while I do that.


python/activator/middleware_interface.py
```diff
@@ -54,9 +54,10 @@ class MiddlewareInterface:
 
     Parameters
     ----------
-    input_repo : `str`
+    central_butler : `lsst.daf.butler.Butler`
         Path to a butler repo containing the calibration and other data needed
-        for processing images as they are received.
+        for processing images as they are received. This butler must be created
+        with the default instrument and skymap assigned.
```
kfindeisen, line 62: Ironically, you didn't update the line that's most relevant to the change.

python/activator/activator.py
```diff
@@ -70,8 +70,15 @@
 
 # Initialize middleware interface; TODO: we'll need one of these per detector.
 repo = f"/tmp/butler-{os.getpid()}"
+central_butler = Butler(calib_repo,
+                        # TODO: How do we get the appropriate instrument name
```
ktlim, line 74: Why is `config_instrument` incorrect?  Or if there needs to be a lookup table, that's OK, as there is a well-defined list of supported instruments.

python/activator/middleware_interface.py
```diff
@@ -54,9 +56,10 @@ class MiddlewareInterface:
 
     Parameters
     ----------
-    input_repo : `str`
+    central_butler : `lsst.daf.butler.Butler`
         Path to a butler repo containing the calibration and other data needed
```
ktlim, line 60: No longer a path, now an object.

tests/test_middleware_interface.py
```diff
@@ -73,12 +73,17 @@ class MiddlewareInterfaceTest(unittest.TestCase):
...
+                                instrument="LSSTCam",
+                                skymap="deepCoadd_skyMap",
+                                writeable=False)
         instrument = "lsst.obs.lsst.LsstCam"
         self.input_data = os.path.join(data_dir, "input_data")
         self.repo = tempfile.TemporaryDirectory()
         butler = Butler(Butler.makeRepo(self.repo.name), writeable=True)
         self.interface = MiddlewareInterface(input_repo, self.input_data, instrument, butler,
+        self.interface = MiddlewareInterface(central_butler, self.input_data, instrument, butler,
                                              prefix="file://")
```
kfindeisen, line 92:
```suggestion
        butler = Butler(Butler.makeRepo(self.repo.name), writeable=True)
        self.interface = MiddlewareInterface(central_butler, self.input_data, instrument, butler,
                                             prefix="file://")
```

python/activator/middleware_interface.py
```diff
@@ -150,11 +155,14 @@ def _init_local_butler(self, butler: Butler):
...
         # structure from an export.yaml file.
-        self.instrument.register(butler.registry)
+        # TODO: Cannot do this until we have a way to only extract specific
+        # calibs, otherwise we get a unique constraint error on import.
+        # self.instrument.register(butler.registry)
         # Refresh butler after configuring it, to ensure all required
         # dimensions are available.
         butler.registry.refresh()
         self.butler = butler
+        self.camera = self.central_butler.get("camera", instrument=self.instrument.getName())
```
ktlim, line 165: Could move to `__init__`, no?

tests/data/make_central_repo.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash -e
+# Create a central repo that contains calibs, refcats, templates for use in
+# tests, making all the files be empty and adjusting the sqlite registry to
+# match.
+
+butler create central_repo
+butler import --export-file export.yaml --transfer copy central_repo/ $AP_VERIFY_CI_HITS2015_DIR/preloaded/
```
kfindeisen, line 7: Note that the `export.yaml` will bit-rot as we make changes to the `ap_verify` dataset. Might it be better to commit the export command instead of the file?

python/activator/middleware_interface.py
```diff
@@ -176,42 +184,122 @@ def prep_butler(self, visit: Visit) -> None:
...
-        #     for collection in export_collections:
-        #         e.saveCollection(collection)
-        #     e.saveDatasets(export_datasets)
-        _log.debug(f"Importing from {visit_dir}")
-        # self.butler.import_(directory=visit_dir, format="yaml", transfer="hardlink")
-        shutil.rmtree(visit_dir, ignore_errors=True)
+
+        with tempfile.NamedTemporaryFile(mode="w+b", suffix=".yaml") as export_file:
+            with self.central_butler.export(filename=export_file.name, format="yaml") as export:
+                # TODO: we may want to refactor Visit to store the boresight
```
ktlim, line 190: I think you did this 😀

python/activator/middleware_interface.py
```diff
@@ -176,42 +184,122 @@ def prep_butler(self, visit: Visit) -> None:
...
+        wcs : `lsst.afw.geom.SkyWcs`
+            Rough WCS for the upcoming visit, to help finding patches.
+        """
+        export.saveDatasets(self.central_butler.registry.queryDatasets("skyMap",
+                                                                       collections="skymaps",
+                                                                       findFirst=True))
+        # Getting only one tract should be safe: we're getting the
+        # tract closest to this detector, so we should be well within
+        # the tract bbox.
+        skymap = self.central_butler.get("skyMap")
```
ktlim, line 260: Also load once in `__init__`?

python/activator/middleware_interface.py
```diff
@@ -176,42 +184,122 @@ def prep_butler(self, visit: Visit) -> None:
...
+        tract = skymap.findTract(center)
+        points = [center]
+        for corner in detector.getCorners(lsst.afw.cameraGeom.PIXELS):
+            points.append(wcs.pixelToSky(corner))
+        patches = tract.findPatchList(points)
+        template_where = f"patch in ({','.join(str(p.sequential_index) for p in patches)}) and tract={tract.tract_id}"
+        # TODO: do we need to have the coadd name used in the pipeline
```
ktlim, line 267: I think having it be extracted from the pipeline is best and configured into the service is next best.

python/activator/middleware_interface.py
```diff
@@ -150,11 +155,14 @@ def _init_local_butler(self, butler: Butler):
         """
         # TODO: Replace registring the instrument with importing a "base" repo
         # structure from an export.yaml file.
-        self.instrument.register(butler.registry)
+        # TODO: Cannot do this until we have a way to only extract specific
+        # calibs, otherwise we get a unique constraint error on import.
+        # self.instrument.register(butler.registry)
```
kfindeisen, line 160: I don't see how calib selectivity is related to the uniqueness constraint (though I agree that not calling `instrument.register` is probably faster than debugging `saveDatasets(element=...)`).

python/activator/middleware_interface.py
```diff
@@ -176,42 +184,122 @@ def prep_butler(self, visit: Visit) -> None:
...
+        export.saveDatasets(self.central_butler.registry.queryDatasets(
+            ...,
+            collections=self.instrument.makeCalibrationCollectionName(),
+            where=calib_where))
+        target_types = {CollectionType.CALIBRATION}
+        for collection in self.central_butler.registry.queryCollections(...,
+                                                                        collectionTypes=target_types):
+            export.saveCollection(collection)
+
+        # TODO: Should probably also get the appropriate chains here, too?
```
ktlim, line 301: I guess if you provide all run collections to the pipeline executor, the chains don't matter.

tests/test_middleware_interface.py
```diff
@@ -109,24 +123,28 @@ def test_init(self):
...
-        with self.assertLogs(self.logger_name, level="INFO") as cm:
-            self.interface.prep_butler(self.next_visit)
-        msg = f"INFO:{self.logger_name}:Preparing Butler for visit '{self.next_visit}'"
-        self.assertEqual(cm.output, [msg])
-        # TODO: Test that we have appropriate refcats?
+        self.interface.prep_butler(self.next_visit)
+        self.assertEqual(self.butler.get('camera',
+                                         instrument=instname,
+                                         collections=[f"{instname}/calib/unbounded"]).getName(), instname)
+        # TODO: check that we got the correct refcat shards. How to do that?
```
ktlim, line 130: For the short term, you could precalculate and hard-code for your known test data.

python/activator/middleware_interface.py
```diff
@@ -176,42 +184,122 @@ def prep_butler(self, visit: Visit) -> None:
...
+        points = [center]
+        for corner in detector.getCorners(lsst.afw.cameraGeom.PIXELS):
+            points.append(wcs.pixelToSky(corner))
+        patches = tract.findPatchList(points)
+        template_where = f"patch in ({','.join(str(p.sequential_index) for p in patches)}) and tract={tract.tract_id}"
+        # TODO: do we need to have the coadd name used in the pipeline
+        # specified as a class kwarg, so that we only load one here?
+        # TODO: alternately, we need to extract it from the pipeline?
+        # TODO: alternately, can we just assume that there is exactly
+        # one coadd in the central butler?
```
kfindeisen, line 271:
```suggestion
        # one coadd type in the central butler?
```

And I think the last is a safe assumption, at least for now. Only if the central repo started including DRP outputs would it be likely to be broken.

python/activator/middleware_interface.py
```diff
@@ -176,42 +184,122 @@ def prep_butler(self, visit: Visit) -> None:
...
+            Export context manager.
+        detector_id : `int`
+            Identifier of the detector to load calibs for.
+        filter : `str`
+            Physical filter name of the upcoming visit.
+        """
+        # TODO: we can't filter by validity range because it's not
+        # supported in queryDatasets yet.
+        calib_where = f"detector={detector_id} and physical_filter='{filter}'"
+        # import ipdb; ipdb.set_trace()
```
kfindeisen, line 291: Rogue debugging line.

ktlim (approved):
Looks good enough.  Will need to be reconciled with the uploader, now that the `visit` definition is changing.  I worry a bit about serialization and `to_dict` with the `lsst.afw.geom` classes, but presumably it works well enough?

python/activator/middleware_interface.py
```diff
@@ -176,42 +184,122 @@ def prep_butler(self, visit: Visit) -> None:
...
+            ...,
+            collections=self.instrument.makeCalibrationCollectionName(),
+            where=calib_where))
+        target_types = {CollectionType.CALIBRATION}
+        for collection in self.central_butler.registry.queryCollections(...,
+                                                                        collectionTypes=target_types):
+            export.saveCollection(collection)
+
+        # TODO: Should probably also get the appropriate chains here, too?
+        # e.g. chain unbounded to calib...
```
kfindeisen, line 302: `calib` is not a chain, it's a calibration collection. It has its own rules for tracing datasets back to their run collections.

In any case, there's no need to do anything for this.

tests/test_middleware_interface.py
```diff
@@ -68,30 +68,37 @@ def fake_file_data(filename, dimensions, instrument):
...
+
 class MiddlewareInterfaceTest(unittest.TestCase):
     """Test the MiddlewareInterface class with faked data.
     """
     def setUp(self):
         data_dir = os.path.join(os.path.abspath(os.path.dirname(__file__)), "data")
         central_repo = os.path.join(data_dir, "central_repo")
         central_butler = Butler(central_repo,
-                                instrument="LSSTCam",
+                                instrument=instname,
```
kfindeisen, line 82: Should this commit ("Use DECam in tests") be merged with the one that added `central_repo`?

tests/test_middleware_interface.py
```diff
@@ -85,10 +85,9 @@ def setUp(self):
                                 writeable=False)
         instrument = "lsst.obs.decam.DarkEnergyCamera"
         self.input_data = os.path.join(data_dir, "input_data")
-        self.repo = tempfile.TemporaryDirectory()
-        instrument = lsst.obs.base.utils.getInstrument(instrument)
-        butler = Butler(Butler.makeRepo(self.repo.name), instrument="LSSTCam", writeable=True)
-        self.interface = MiddlewareInterface(central_butler, self.input_data, instrument, butler,
+        repo = tempfile.TemporaryDirectory()
+        self.butler = Butler(Butler.makeRepo(repo.name), writeable=True)
```
kfindeisen: `self.local_butler`, just to be unambiguous?

python/activator/activator.py
kfindeisen: This has to do with some UIs requiring a class name and others an instrument short name. As mentioned on Slack, I would just create a global `Instrument` object and then use that for the translation.

tests/test_middleware_interface.py
```diff
@@ -85,8 +85,9 @@ def setUp(self):
                                 writeable=False)
         instrument = "lsst.obs.decam.DarkEnergyCamera"
         self.input_data = os.path.join(data_dir, "input_data")
-        repo = tempfile.TemporaryDirectory()
-        self.butler = Butler(Butler.makeRepo(repo.name), writeable=True)
+        # Have to preserve the tempdir, so that it doesn't get cleaned up.
+        self.repo = tempfile.TemporaryDirectory()
+        self.butler = Butler(Butler.makeRepo(self.repo.name), writeable=True)
```
kfindeisen, line 90: Merge "Fix test tempdir bug" with "Save the local butler to query it in tests"?

tests/test_middleware_interface.py
```diff
@@ -110,9 +110,11 @@ def test_init(self):
         #   collections, etc?
         # * On init, is the local butler repo purely in memory?
 
+        # TODO: this no longer works because we cannot register an instrument
+        # and also import an export that contains that instrument.
         # Check that the butler instance is properly configured.
-        instruments = list(self.butler.registry.queryDimensionRecords("instrument"))
-        self.assertEqual(instname, instruments[0].name)
+        # instruments = list(self.butler.registry.queryDimensionRecords("instrument"))
+        # self.assertEqual(instname, instruments[0].name)
```
kfindeisen, line 117: I don't understand what is accomplished by commenting out these assertions. The `queryDimensionRecords` call should not attempt to modify the registry...

kfindeisen (approved):
Looks good, most of my commends are tidiness-related.

python/activator/middleware_interface.py
kfindeisen: Providing runs to the pipeline is going to be impractical in the long run, because they depend on where exactly the data came from (and this goes for test data as well). While they're hard to export because you need exact definitions, chains make it *much* easier to program the pipeline parts against a specific interface.

python/activator/activator.py
parejkoj: I'm going to leave this question to the "reintegrate activator and interface" ticket (that I guess doesn't exist?), if that's ok? We could create an `Instrument` here, but we'd still have to change the dict in `uploader.py` to have full class names, I think? And we'd also have to expand the eups table to have all the correct obs packages. I feel like that's enough moving parts that it should be a separate ticket.

kfindeisen: I don't see any reason why we'd have to change `uploader.py`; we'd just need to make sure _this file_ is always keeping straight when it wants the class name and when it wants the short name.

python/activator/middleware_interface.py
parejkoj: We absolutely want to program assuming a default set of chains. That's what they're for. I think I've got it now anyway.

tests/test_middleware_interface.py
parejkoj: Part of the question was "what *are* the correct refcat shards?" See DM-34112 for further investigation.

At least I have some kind of test in here now.

tests/test_middleware_interface.py
parejkoj: I think this is a result of me trying to make more atomic commits: it's fixed in the final version somewhere...

tests/data/make_central_repo.sh
parejkoj: I also had to tweak the export file to include the CHAINED collections that we want. I'll add a note here about that, and the command I used to run it.

python/activator/middleware_interface.py
parejkoj: If we process the same detector in the same MiddlewareInterface, it will fail on `prep_butler` because all the calibs already exist, even independent of the `camera` dataset or the instrument registration.

python/activator/middleware_interface.py
parejkoj: In this case, we did want to get the `refcats` chain at least.

tests/test_middleware_interface.py
parejkoj: Could be? I was trying to be more specific in my commits as you wanted, and this one was mostly atomic.

tests/test_middleware_interface.py
parejkoj: It ends up making a bunch of things rather longer though: we use it quite a bit.

tests/test_middleware_interface.py
parejkoj: We can't make this check after `init` because nothing exists after `init` now, because we can't register the instrument before the `import`.

tests/data/make_central_repo.sh
kfindeisen: Yikes. And I guess there's no time to test a file that includes the chains from the beginning...

python/activator/middleware_interface.py
kfindeisen: I still don't see why "extract specific calibs" matters; even with that capability, you'd still risk double-defining data IDs on later imports.

parejkoj:
I'll note that the fake repo is about 3MB in size, almost all of which is the gen3 registry sqlite file. I don't think that's enough to really care about how much size it adds to this repo, but we could strip it down somewhat by removing all the unnecessary calib and refcat references (since we only have 4 fake detectors here anyway)

tests/test_middleware_interface.py
kfindeisen: I think this one is subatomic, if the previous `Butler` call wouldn't actually have worked.

python/activator/middleware_interface.py
parejkoj: If we can be more selective, we can just export the calibs we don't have, so we don't have the double-definition problem.

kfindeisen: Got it. Then I suggest rewording to something like "until we have a way to only extract calibs we don't already have". I thought you were talking about the validities.

PR #15, opened 2022-03-17
DM-33970: Use real images in upload.py

kfindeisen:
This PR does a lot of refactoring of `upload.py` to make it easier to edit and more flexible with its inputs, then adds partial support for "observing" preloaded raw files. Because some of the raw metadata is hardcoded into the `get_samples` function, this script will break if any other raws are uploaded to `rubin-prompt-proto-unobserved`.

Because of the heavy refactoring on the first part of the ticket, I recommend reviewing this PR one or a few commits at a time; the net diff combines all the changes into one big rewrite.

### DIFF
--- a/python/tester/upload.py
+++ b/python/tester/upload.py
@@ -171,48 +171,10 @@ def main():
 
     if raw_pool:
         _log.info(f"Observing real raw files from {instrument}.")
-        for i, true_group in enumerate(itertools.islice(itertools.cycle(raw_pool), n_groups)):
-            group = last_group + i + 1
-            _log.debug(f"Processing group {group} from unobserved {true_group}...")
-            # snap_dict maps snap_id to {visit: blob}
-            snap_dict = {}
-            # Copy all the visit-blob dictionaries under each snap_id,
-            # replacing the (immutable) Visit objects to point to group
-            # instead of true_group.
-            for snap_id, old_visits in raw_pool[true_group].items():
-                snap_dict[snap_id] = {splice_group(true_visit, group): blob
-                                      for true_visit, blob in old_visits.items()}
-            # Gather all the Visit objects found in snap_dict, merging
-            # duplicates for different snaps of the same detector.
-            visit_infos = {info for det_dict in snap_dict.values() for info in det_dict}
-
-            # TODO: may be cleaner to use a functor object than to depend on
-            # closures for the bucket and data.
-            def upload_from_pool(visit, snap_id):
-                src_blob = snap_dict[snap_id][visit]
-                exposure_id = make_exposure_id(visit.instrument, visit.group, snap_id)
-                filename = raw_path(visit.instrument, visit.detector, visit.group, snap_id,
-                                    exposure_id, visit.filter)
-                src_bucket.copy_blob(src_blob, dest_bucket, new_name=filename)
-            process_group(publisher, visit_infos, upload_from_pool)
-            _log.info("Slewing to next group")
-            time.sleep(SLEW_INTERVAL)
+        upload_from_raws(publisher, instrument, raw_pool, src_bucket, dest_bucket, n_groups, last_group + 1)
     else:
         _log.info(f"No raw files found for {instrument}, generating dummy files instead.")
-        for i in range(n_groups):
-            group = last_group + i + 1
-            visit_infos = make_random_visits(instrument, group)
-
-            # TODO: may be cleaner to use a functor object than to depend on
-            # closures for the bucket and data.
-            def upload_dummy(visit, snap_id):
-                exposure_id = make_exposure_id(visit.instrument, visit.group, snap_id)
-                filename = raw_path(visit.instrument, visit.detector, visit.group, snap_id,
-                                    exposure_id, visit.filter)
-                dest_bucket.blob(filename).upload_from_string("Test")
-            process_group(publisher, visit_infos, upload_dummy)
-            _log.info("Slewing to next group")
-            time.sleep(SLEW_INTERVAL)
+        upload_from_random(publisher, instrument, dest_bucket, n_groups, last_group + 1)
 
 
 def get_last_group(storage_client, instrument, date):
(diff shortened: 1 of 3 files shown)


python/tester/upload.py
```diff
@@ -51,18 +51,46 @@ def raw_path(instrument, detector, group, snap, exposure_id, filter):
...
+def process_group(publisher, bucket, visit_infos):
+    """Simulate the observation of a single on-sky pointing.
+
+    Parameters
+    ----------
+    publisher : `google.cloud.pubsub_v1.PublisherClient`
+        The client that posts ``next_visit`` messages.
+    bucket : `google.cloud.storage.Bucket`
+        The bucket to which to transfer the raws, once observed.
+    visit_infos : `set` [`activator.Visit`]
```
parejkoj, line 70: I'm a bit nervous calling these "visit_infos", since `VisitInfo` is a different thing that doesn't align with this very well (though maybe we should make it do so?). Why not just `visits`?

python/tester/upload.py
```diff
@@ -165,7 +164,15 @@ def main():
             Visit(instrument, detector, group, INSTRUMENTS[instrument].n_snaps, filter, ra, dec, kind)
             for detector in range(INSTRUMENTS[instrument].n_detectors)
         }
-        process_group(publisher, bucket, visit_infos)
+
+        # TODO: may be cleaner to use a functor object than to depend on
+        # closures for the bucket and data.
+        def upload_dummy(visit, snap_id):
+            exposure_id = make_exposure_id(visit.instrument, visit.group, snap_id)
+            fname = raw_path(visit.instrument, visit.detector, visit.group, snap_id,
```
parejkoj: fname->filename?

ktlim (approved):
I'm not exactly sure why the uploader needs to be so dynamic, but I'll let John figure that out.  The rest looks fine.  At some point we need to straighten out what's a group and what's a visit, but the distinction only matters in specialized cases.

kfindeisen:
> I'm not exactly sure why the uploader needs to be so dynamic

Because, for me, modifying the existing code was easier than trying to figure out brand-new behavior (e.g., CLI) from scratch.

python/tester/upload.py
```diff
@@ -84,34 +153,222 @@ def main():
...
     publisher = pubsub_v1.PublisherClient(credentials=credentials,
                                           batch_settings=batch_settings)
 
+    last_group = get_last_group(storage_client, instrument, date)
+    _log.info(f"Last group {last_group}")
+
+    src_bucket = storage_client.bucket("rubin-prompt-proto-unobserved")
+    raw_pool = get_samples(src_bucket, instrument)
+
+    if raw_pool:
```
parejkoj, line 169: I kind of want the two parts of this `if` to be separate functions: `handle_real_raws` and `handle_random_raws` or something like that?

python/tester/upload.py
```diff
@@ -84,34 +153,222 @@ def main():
...
+            time.sleep(SLEW_INTERVAL)
+    else:
+        _log.info(f"No raw files found for {instrument}, generating dummy files instead.")
+        for i in range(n_groups):
+            group = last_group + i + 1
+            visit_infos = make_random_visits(instrument, group)
+
+            # TODO: may be cleaner to use a functor object than to depend on
+            # closures for the bucket and data.
+            def upload_dummy(visit, snap_id):
```
parejkoj, line 200: Maybe `upload_faked_files` instead of dummy? Or does that not even upload files at all, so just `upload_faked_visits_no_files` (which is too long, but...)

python/tester/upload.py
```diff
@@ -39,38 +60,86 @@ class Instrument:
...
+    # Assume group/snaps is shared among all visit_infos
+    for info in visit_infos:
+        group = info.group
+        n_snaps = info.snaps
+        break
+    else:
+        _log.info("No observations to make; aborting.")
+        return
+
+    send_next_visit(publisher, group, visit_infos)
```
ktlim, line 87: It's a bit aggressive to be immediately "taking an image" after sending next_visit.  In theory we're supposed to have 20 seconds prior to the exposure being taken, so we could add a 20 sec sleep here.

python/tester/upload.py
```diff
@@ -84,34 +153,222 @@ def main():
...
+        _log.info(f"Observing real raw files from {instrument}.")
+        for i, true_group in enumerate(itertools.islice(itertools.cycle(raw_pool), n_groups)):
+            group = last_group + i + 1
+            _log.debug(f"Processing group {group} from unobserved {true_group}...")
+            # snap_dict maps snap_id to {visit: blob}
+            snap_dict = {}
+            for snap_id, old_visits in raw_pool[true_group].items():
+                snap_dict[snap_id] = {splice_group(true_visit, group): blob
+                                      for true_visit, blob in old_visits.items()}
+            visit_infos = {info for det_dict in snap_dict.values() for info in det_dict}
```
parejkoj, line 179: The above `for` loop and dict comps seem difficult to read for me, but I'm not sure that there's anything to be done about it. Maybe at least a comment explaining what this block is supposed to do?

python/tester/upload.py
```diff
@@ -84,34 +153,222 @@ def main():
...
+    """Identify a group number that will not collide with any previous groups.
+
+    Parameters
+    ----------
+    storage_client : `google.cloud.storage.Client`
+        A Google Cloud Storage object pointing to the active project.
+    instrument : `str`
+        The short name of the active instrument.
+    date : `str`
+        The current date in YYYYMMDD format.
```
parejkoj, line 220: Not a datetime or astropy date?

python/tester/upload.py
ktlim: Oh, except that unrealistically gives a 20 second extra delay between groups.  We'd have to do something more complicated with asynchronous or threaded routines.

kfindeisen: A preexisting bug, but I'm happy to add fix it on this ticket.

python/tester/upload.py
kfindeisen: Because they're not visits either (in any of the senses of the word)! 😭 

A name like "<something>_info" feels natural to me, because these are blocks of metadata rather than identifiers (like `visit` in the Butler context). I'm willing to be flexible on the first part, though.

python/tester/upload.py
kfindeisen: This does, in fact, upload a file -- most of the contents of `-main` were uploaded using this code (pre-refactor, of course).

python/tester/upload.py
kfindeisen: No. This is part of the original group generation code, and I don't want to touch it because then I'd need to come up with a different scheme for avoiding collisions between uploaded files.

python/tester/upload.py
kfindeisen: Agreed on Slack to leave the behavior as-is.

PR #16, opened 2022-03-18
DM-34123: Use hashable Visit in upload.py

ebellm:


python/tester/upload.py
```diff
@@ -10,8 +10,6 @@
 import time
 from visit import Visit
```
ktlim, line 11: Maybe call this `VisitMessage` to be more clear?

ktlim (approved):
LGTM.

PR #17, opened 2022-03-21
DM-34132: Use shared visit.py with pure python types

ebellm:


python/activator/visit.py
```diff
@@ -11,6 +10,7 @@ class Visit:
     group: str
     snaps: int
     filter: str
-    boresight_center: lsst.geom.SpherePoint
-    orientation: lsst.geom.Angle
+    ra: float
+    dec: float
+    rot: float
```
kfindeisen, line 15: I suggest adding a comment that these are all in degrees.

python/activator/visit.py
```diff
@@ -1,7 +1,6 @@
 __all__ = ["Visit"]
 
 from dataclasses import dataclass
-import lsst.geom
 
 
 @dataclass(frozen=True)
```
kfindeisen, line 6: Please add a comment warning that `Visit`'s members should be builtins, and must be both hashable and JSON-persistable. Otherwise we'll just make the same mistake again.

python/tester/upload.py
```diff
@@ -299,10 +299,9 @@ def get_samples(bucket, instrument):
                       group=group,
                       snaps=INSTRUMENTS[instrument].n_snaps,
                       filter=parsed.group('filter'),
-                      boresight_center=SpherePoint(hsc_metadata[exposure_id]["ra"],
-                                                   hsc_metadata[exposure_id]["dec"],
-                                                   degrees),
-                      orientation=hsc_metadata[exposure_id]["rot"] * degrees,
+                      ra=hsc_metadata[exposure_id]["ra"],
+                      dec=hsc_metadata[exposure_id]["dec"],
+                      rot=hsc_metadata[exposure_id]["rot"],
```
kfindeisen, line 304: You missed one of the `Visit` constructor calls, around line 250.

kfindeisen (changes requested):
There's one visible bug (an out-of-date call to `Visit()`). Can you test `upload.py` in the cloud, or should I try to do it?

kfindeisen (approved):

ebellm:
@kfindeisen, can you take another look? I confirmed that upload.py is working, with HSC as well as a simulated instrument.

PR #19, opened 2022-03-21
DM-34017: Use the butler API and mock data for run_pipeline tests

ebellm:


python/activator/middleware_interface.py
```diff
@@ -75,22 +72,35 @@ class MiddlewareInterface:
...
-        self.calibration_collection = f"{instrument}/calib"
+        # TODO: we probably want to be able to configure this per-instrument?
+        # TODO: would be nice to remap ap_pipe/pipelines to use getName convention
+        camera_paths = {"DECam": "DarkEnergyCamera", "HSC": "HyperSuprimeCam"}
+        try:
+            camera_path = camera_paths[self.instrument.getName()]
+        except KeyError:
+            raise NotImplementedError("No ApPipe.yaml defined for camera {}" % self.instrument.getName())
+        ap_pipeline_file = os.path.join(lsst.utils.getPackageDir('ap_pipe'),
+                                        "pipelines/{}/ApPipe.yaml".format(camera_path))
```
ktlim, line 101:
```suggestion
        ap_pipeline_file = os.path.join(lsst.utils.getPackageDir('ap_pipe'),
                                        "pipelines", camera_path, "ApPipe.yaml")
```

python/activator/middleware_interface.py
```diff
@@ -282,21 +322,13 @@ def run_pipeline(self, visit: Visit, snaps: set) -> None:
...
-            "-p",
-            pipeline,
-            "-i",
-            f"{self.instrument}/raw/all",
-        ]
-        _log.debug("pipetask command line: %s", cmd)
-        # subprocess.run(cmd, check=True)
+        where = f"detector={visit.detector} and exposure in ({','.join(str(x) for x in snaps)})"
+        pipeline = SimplePipelineExecutor.from_pipeline(self.pipeline, where=where, butler=self.butler)
+        _log.info(f"Running pipeline {self.pipeline} on visit '{visit}', snaps {snaps}")
```
kfindeisen, line 327: Does this log message still work when `self.pipeline` is a `Pipeline` object instead of a filename?

python/activator/middleware_interface.py
```diff
@@ -251,6 +264,33 @@ def _export_calibs(self, export, detector_id, filter):
...
+        # The below is taken from SimplePipelineExecutor.prep_butler.
+        output_run = f"{self.output_collection}/{self.instrument.makeCollectionTimestamp()}"
+        self.butler.registry.registerCollection(self.instrument.makeDefaultRawIngestRunName(),
+                                                CollectionType.RUN)
+        self.butler.registry.registerCollection(output_run, CollectionType.RUN)
+        self.butler.registry.registerCollection(self.output_collection, CollectionType.CHAINED)
+        collections = [self.calibration_collection,
+                       self.instrument.makeDefaultRawIngestRunName(),
+                       'refcats',
+                       output_run]
```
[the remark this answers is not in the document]
kfindeisen: It might be a good unit test for `_prep_butler` [sic] to check if the test input data can all be found in `self.output_collection` with e.g. `butler.datasetExists`. I think such a test wouldn't be affected by the fact that all the files are empty.

tests/test_middleware_interface.py
```diff
@@ -188,6 +195,8 @@ def test_ingest_image(self):
             datasets = list(self.butler.registry.queryDatasets('raw',
                                                                collections=[f'{instname}/raw/all']))
             self.assertEqual(datasets[0].dataId, data_id)
+            # TODO: I think we also need to check that the visits are defined?
+            # Or should that happen in prep_butler?
```
kfindeisen, line 199: We _should_ define visits ASAP, but it's impossible to do it before `exposure` dimension records are defined, and at present it's very hard to do that before raws are ingested. See my note in the sprint summary document.

tests/test_middleware_interface.py
```diff
@@ -119,8 +121,13 @@ def test_init(self):
         # TODO: this no longer works because we cannot register an instrument
         # and also import an export that contains that instrument.
         # Check that the butler instance is properly configured.
-        # instruments = list(self.butler.registry.queryDimensionRecords("instrument"))  # noqa: W505
-        # self.assertEqual(instname, instruments[0].name)
+        instruments = list(self.butler.registry.queryDimensionRecords("instrument"))  # noqa: W505
```
kfindeisen, line 124: The `noqa` is no longer required now that this is not a comment.

tests/test_middleware_interface.py
```diff
@@ -219,7 +228,17 @@ def test_ingest_image_fails_missing_file(self):
...
-        self.assertEqual(cm.output, [msg])
+        # Can't run an actual pipeline because raw/calib/refcat/template data are all zeroed out.
+        with unittest.mock.patch('activator.middleware_interface.SimplePipelineExecutor') as mock_executor:
+            self.interface.run_pipeline(self.next_visit, {1})
+            mock_executor.from_pipeline.assert_called_once_with(
+                self.interface.pipeline,
+                where=f"detector={self.next_visit.detector} and exposure in (1)",
+                butler=self.interface.butler
+            )
+            pipeline = mock_executor.from_pipeline(
```
kfindeisen, line 239: These are some unfortunately chosen variable names, in hindsight. I'd suggest calling `mock_executor` either `mock_class` or `mock_executor_class`, and renaming `pipeline` to `mock_executor` or `mock_executor_instance`.

python/activator/middleware_interface.py
```diff
@@ -282,21 +322,13 @@ def run_pipeline(self, visit: Visit, snaps: set) -> None:
...
-            self.butler,
-            "-p",
-            pipeline,
-            "-i",
-            f"{self.instrument}/raw/all",
-        ]
-        _log.debug("pipetask command line: %s", cmd)
-        # subprocess.run(cmd, check=True)
+        where = f"detector={visit.detector} and exposure in ({','.join(str(x) for x in snaps)})"
+        pipeline = SimplePipelineExecutor.from_pipeline(self.pipeline, where=where, butler=self.butler)
```
kfindeisen, line 326: `pipeline` is actually a `SimplePipelineExecutor` object... maybe call it `executor`?

ups/prompt_prototype.table
```diff
@@ -7,10 +7,13 @@ setupRequired(sconsUtils)
 setupRequired(utils)         # Used by scripts in bin.src
 
 # Used by middleware_interface module.
+setupRequired(ap_pipe)
```
kfindeisen, line 10: Worth adding a comment that this line is required because of the hardcoded pipeline references, not from any code.

kfindeisen (approved):
Looks good, though the current collection handling will cause problems in integration.

python/activator/middleware_interface.py
```diff
@@ -75,22 +72,35 @@ class MiddlewareInterface:
...
+        # TODO: would be nice to remap ap_pipe/pipelines to use getName convention
+        camera_paths = {"DECam": "DarkEnergyCamera", "HSC": "HyperSuprimeCam"}
+        try:
+            camera_path = camera_paths[self.instrument.getName()]
+        except KeyError:
+            raise NotImplementedError("No ApPipe.yaml defined for camera {}" % self.instrument.getName())
+        ap_pipeline_file = os.path.join(lsst.utils.getPackageDir('ap_pipe'),
+                                        "pipelines/{}/ApPipe.yaml".format(camera_path))
+        self.pipeline = lsst.pipe.base.Pipeline.fromFile(ap_pipeline_file)
+        self.pipeline.addConfigOverride("diaPipe", "apdb.db_url", "postgresql://postgres@localhost/postgres")
```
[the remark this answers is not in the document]
ebellm: I guess I will add apdb IP as an environment variable in Cloud Run...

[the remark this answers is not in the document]
kfindeisen: Should we do that for both databases? Note that the `butler.yaml` files will need a hardcoded value regardless.

python/activator/middleware_interface.py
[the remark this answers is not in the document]
kfindeisen: `output_collection` is a chained collection that includes all its inputs. This is the standard Gen 3 repository structure for doing pipeline runs.

python/activator/middleware_interface.py
```diff
@@ -75,22 +72,35 @@ class MiddlewareInterface:
...
+        # TODO: we probably want to be able to configure this per-instrument?
+        # TODO: would be nice to remap ap_pipe/pipelines to use getName convention
+        camera_paths = {"DECam": "DarkEnergyCamera", "HSC": "HyperSuprimeCam"}
+        try:
+            camera_path = camera_paths[self.instrument.getName()]
+        except KeyError:
+            raise NotImplementedError("No ApPipe.yaml defined for camera {}" % self.instrument.getName())
+        ap_pipeline_file = os.path.join(lsst.utils.getPackageDir('ap_pipe'),
+                                        "pipelines/{}/ApPipe.yaml".format(camera_path))
+        self.pipeline = lsst.pipe.base.Pipeline.fromFile(ap_pipeline_file)
```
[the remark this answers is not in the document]
parejkoj: I don't like keeping the PIPELINE_MAP approach, since we don't know why a given pipeline would be chosen right now (i.e. until we know what the trigger for switching pipelines is, it's hard to pick how to make the switch). Maybe we could move this pipeline definition into `prep_butler`, since if we do switch pipelines, it would probably happen in there?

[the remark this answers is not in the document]
kfindeisen: @ktlim assumed it would be keyed to the `kind` field in the `next_visit` message. I think sticking to that assumption is good enough for now, since it means that `MiddlewareInterface` is not responsible for the decision.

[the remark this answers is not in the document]
parejkoj: Do we have any facility for different "kind"s of pipelines in ap_pipe?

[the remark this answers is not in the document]
ebellm: They would be just be different `ap_pipe/pipelines/.yaml` pipelines.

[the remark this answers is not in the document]
kfindeisen: All I'm asking is to not remove the support that was already implemented.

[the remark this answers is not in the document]
parejkoj: I think what I'm trying to say is that it's not "support" if it doesn't actually do anything right now (we don't have and don't want ap pipelines for calib files), and doesn't look like what we'll actually need in the end (we need to figure out what the ap_pipe part of this will look like, too).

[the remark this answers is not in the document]
kfindeisen: Please do not try to redesign `ap_pipe` based on implementation details of `MiddlewareInterface`! It's very hard to maintain code where every component needs to know everything about what every other component does.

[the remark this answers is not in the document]
parejkoj: I agree completely! I'm not saying we should redesign ap_pipe at all. I'm saying we'll have to think about that when it comes time to implement these things, and what was in here previously is not what we actually want.

python/activator/middleware_interface.py
```diff
@@ -325,8 +325,9 @@ def run_pipeline(self, visit: Visit, snaps: set) -> None:
         where = f"detector={visit.detector} and exposure in ({','.join(str(x) for x in snaps)})"
         pipeline = SimplePipelineExecutor.from_pipeline(self.pipeline, where=where, butler=self.butler)
         _log.info(f"Running pipeline {self.pipeline} on visit '{visit}', snaps {snaps}")
-        # All dataset types should have been pre-registered.
-        result = pipeline.run(register_dataset_types=False)
+        # If this is a fresh (local) repo, then types like calexp,
+        # *Diff_diaSrcTable, etc. have not been registered.
+        result = pipeline.run(register_dataset_types=True)
```
parejkoj, line 330: Can we do that type registration in prep_butler, possibly as part of `_prep_collections`?

kfindeisen: That would require a list of every connection used by every task in the pipeline.

parejkoj: Doesn't the QuantumGraph know that, though?

python/activator/middleware_interface.py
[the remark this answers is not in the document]
ktlim: What I was thinking of was that `kind` would trigger different pipeline YAML files (e.g. ones in https://github.com/lsst/cp_pipe/tree/main/pipelines) as well as possibly different code in `prep_butler()`.  You shouldn't assume that everything being run by this framework is AP-written.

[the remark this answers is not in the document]
parejkoj: Yes, but we need to decide how we'll actually go about that decision process (e.g. if we're trying to load packages from elsewhere, what other packages need to go into that dictionary?). Just saying "there will be a dict" seems too simple and misleading.

tests/test_middleware_interface.py
```diff
@@ -76,8 +75,22 @@ def fake_file_data(filename, dimensions, instrument):
 class MiddlewareInterfaceTest(unittest.TestCase):
     """Test the MiddlewareInterface class with faked data.
     """
+    @classmethod
+    def setUpClass(cls):
+        cls.env_patcher = unittest.mock.patch.dict(os.environ,
+                                                   {"IP_APDB": "localhost"})
```
kfindeisen, line 81: Why is this patch needed? Is `make_pgpass` somehow being called in the test environment? It wasn't before, and shouldn't be. 😨

ebellm: No, `middleware_interface` needs the env variable when it sets up the pipeline (currently at https://github.com/lsst-dm/prompt_prototype/blob/501d70450f88f816cff9c39ffbc4c7f548092ccc/python/activator/middleware_interface.py#L112).  If @parejkoj moves the pipeline initialization we might be able to patch only on the relevant tests (`prep_butler`) and could avoid the env_patching which is needed when the import happens at the top level of `middleware_interface.py`. 

Apologies I didn't think of moving it earlier.

python/activator/middleware_interface.py
```diff
@@ -75,22 +79,37 @@ class MiddlewareInterface:
         appropriate for use in the Google Cloud environment; typically only
         change this when running local tests.
     """
+
     def __init__(self, central_butler: Butler, image_bucket: str, instrument: str,
                  butler: Butler,
-                 prefix: str = "gs://"):
+                 prefix: str = "s3://"):
```
kfindeisen, line 85: This change is no longer necessary after the last weekly, and I've removed S3 support from the Cloud service.

parejkoj:
@kfindeisen : please have another (final?) look. I'll try to combine commits somewhat post-review

python/activator/middleware_interface.py
```diff
@@ -282,20 +278,55 @@ def _prep_collections(self):
...
-        self.butler.registry.registerCollection(output_run, CollectionType.RUN)
+        self.butler.registry.registerCollection(self.output_run, CollectionType.RUN)
         self.butler.registry.registerCollection(self.output_collection, CollectionType.CHAINED)
-        collections = [self.calibration_collection,
+        collections = [self.instrument.makeCollectionName("defaults"),
+                       self.instrument.makeCalibrationCollectionName(),
                        self.instrument.makeDefaultRawIngestRunName(),
                        'refcats',
-                       output_run]
+                       self.output_run]
```
kfindeisen, line 290: The point of using a defaults collection would be that it already includes the calibration and refcats collections, and moves the responsibility for getting it right out of the code and into the dataset.
```suggestion
        collections = [self.instrument.makeCollectionName("defaults"),
                       self.instrument.makeDefaultRawIngestRunName(),
                       self.output_run]
```

python/activator/middleware_interface.py
```diff
@@ -282,20 +278,55 @@ def _prep_collections(self):
         #             'refcats'],
         #     output=self.output_collection)
         # The below is taken from SimplePipelineExecutor.prep_butler.
-        output_run = f"{self.output_collection}/{self.instrument.makeCollectionTimestamp()}"
+        self.output_run = f"{self.output_collection}/{self.instrument.makeCollectionTimestamp()}"
```
kfindeisen, line 281: If hanging onto this at the object level is part of the workaround for DM-34202, can you add a comment to that effect? I would have thought that you could just use `self.output_collection` once it was properly set up.

python/activator/middleware_interface.py
```diff
@@ -282,20 +278,55 @@ def _prep_collections(self):
...
+        """
+        # TODO: we probably want to be able to configure this per-instrument?
+        # TODO: DM-33453 will remap ap_pipe/pipelines to use getName convention
+        camera_paths = {"DECam": "DarkEnergyCamera", "HSC": "HyperSuprimeCam"}
+        try:
+            camera_path = camera_paths[self.instrument.getName()]
+        except KeyError:
+            raise RuntimeError(f"No ApPipe.yaml defined for camera {self.instrument.getName()}")
+        # TODO: DECam test data uses `panstarrs/gaia` for the refcat names.
+        yaml_file = "ApPipe.yaml" if self.instrument.getName() != "DECam" else "ApPipe_hits2015-3.yaml"
```
kfindeisen, line 322: [Umm... no](https://github.com/lsst/ap_pipe/blob/main/pipelines/DarkEnergyCamera/ApPipe.yaml). `ApPipe_hits2015-3.yaml` is of historical interest and probably won't work with the current Stack.

If the goal is to work around the quirks of `ap_verify` datasets, please configure the pipeline directly. Aside from the refcats, this should just be a matter of setting [the parameters](https://github.com/lsst/ap_pipe/blob/main/pipelines/ApPipe.yaml#L14-L28). Though I wonder if it would be more expedient to just copy the datasets' `ApPipe.yaml` files for now.

tests/test_middleware_interface.py
```diff
@@ -76,8 +75,22 @@ def fake_file_data(filename, dimensions, instrument):
...
+        super().setUpClass()
+
+    @classmethod
+    def tearDownClass(cls):
+        super().tearDownClass()
+
+        cls.env_patcher.stop()
 
     def setUp(self):
+        from activator.middleware_interface import MiddlewareInterface
```
kfindeisen, line 93: Why does this need to go here?

tests/test_middleware_interface.py
```diff
@@ -145,26 +158,48 @@ def test_prep_butler(self):
...
-        expected_shards = [157401, 157405, 157407]
+        expected_shards = [157394, 157401, 157405]
         self.assertEqual(expected_shards, [x['htm7'] for x in loaded_shards])
-        # check that we got appropriate calibs.
+        # Check that the right calibs are in the chained output collection.
         try:
             self.butler.datasetExists('cpBias', detector=56, instrument='DECam',
                                       collections="DECam/calib/20150218T000000Z")
+            # TODO: Have to use the exact calib collection, because we don't
+            # have validity ranges for data.
```
kfindeisen, line 179:
```suggestion
            # TODO: Have to use the exact run, because we can't
            # query by validity range.
```
Please don't use "calib collection" to refer to something that's not a calibration collection; it's confusing. Same for the comment below.

tests/test_middleware_interface.py
```diff
@@ -145,26 +158,48 @@ def test_prep_butler(self):
...
+
+        # Check that we configured the right pipeline.
+        self.assertEquals(self.interface.pipeline._pipelineIR.description,
+                          "End to end AP pipeline for the repo in /project/mrawls/hits2015-3")
+
+        # Check that the right templates are in the chained output collection.
+        for patch in (464, 465, 509, 510):
+            self.butler.datasetExists('deepCoadd', tract=0, patch=patch, band="g",
+                                      # TODO: we shouldn't need skymap here?
+                                      skymap="deepCoadd_skyMap",
```
kfindeisen, line 201: Is this a "just in case", or is the default inference actually not working?

tests/test_middleware_interface.py
```diff
@@ -228,17 +263,21 @@ def test_ingest_image_fails_missing_file(self):
...
+        filename = "fakeRawImage.fits"
+        filepath = os.path.join(self.input_data, filename)
+        data_id, file_data = fake_file_data(filepath,
+                                            self.butler.dimensions,
+                                            self.interface.instrument)
+        with unittest.mock.patch.object(self.interface.rawIngestTask, "extractMetadata") as mock:
+            mock.return_value = file_data
+            self.interface.ingest_image(filename)
+
+        with unittest.mock.patch("activator.middleware_interface.SimplePipelineExecutor.run") as run:
```
kfindeisen, line 281: `mock_run`? I'd like to keep straight which things are mocks and which are real.

tests/test_middleware_interface.py
```diff
@@ -228,17 +263,21 @@ def test_ingest_image_fails_missing_file(self):
...
             self.interface.run_pipeline(self.next_visit, {1})
-            mock_executor.from_pipeline.assert_called_once_with(
-                self.interface.pipeline,
-                where=f"detector={self.next_visit.detector} and exposure in (1)",
-                butler=self.interface.butler
-            )
```
kfindeisen: I think this is a useful test (with maybe some more use of `mock.ANY`), since the restriction to a specific pipeline and data ID is a key part of this method's purpose. It seems a shame to remove it.

python/activator/middleware_interface.py
```diff
@@ -328,12 +359,17 @@ def run_pipeline(self, visit: Visit, snaps: set) -> None:
...
         """
+        # TODO: can we move this from_pipeline call to prep_butler?
         where = f"detector={visit.detector} and exposure in ({','.join(str(x) for x in snaps)})"
-        pipeline = SimplePipelineExecutor.from_pipeline(self.pipeline, where=where, butler=self.butler)
+        executor = SimplePipelineExecutor.from_pipeline(self.pipeline, where=where, butler=self.butler)
+        # TODO DM-34202: hack around a middleware bug.
+        executor.butler = Butler(butler=self.butler,
+                                 collections=[self.output_collection],
+                                 run=self.output_run)
         _log.info(f"Running pipeline {self.pipeline} on visit '{visit}', snaps {snaps}")
```
kfindeisen, line 369: This is a regression of the previous issue where it would put a detailed pipeline dump in the log. I think we're better off having no log message at all.

kfindeisen (approved):
Looks good, but please make sure the tests pass (the flake8 error should have appeared locally), and please avoid special-casing the unit tests. There's not much point to having them if they're free to make different assumptions from the prototype environment.

python/activator/middleware_interface.py
```diff
@@ -128,7 +118,7 @@ def _init_local_butler(self, butler: Butler):
         # self.instrument.register(butler.registry)
 
         # Refresh butler after configuring it, to ensure all required
-        # dimensions are available.
+        # collections are available.
```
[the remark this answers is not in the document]
parejkoj: As it stands, init_butler has nothing to do, since we have to put off all the work for prep_butler right now. We can leave it around while we wait for middleware fixes, etc. that will let us do more things here, but I've added a note to the method docs about whether it's worth keeping the method.

python/activator/middleware_interface.py
```diff
@@ -175,12 +166,17 @@ def prep_butler(self, visit: Visit) -> None:
...
                 export.saveCollection("refcats")
+                export.saveCollection("templates")
+                export.saveCollection(self.instrument.makeCollectionName("defaults"))
 
             self.butler.import_(filename=export_file.name,
                                 directory=self.central_butler.datastore.root,
                                 transfer="copy")
 
+        # Ensure all components of the instrument are registered.
+        self.instrument.register(self.butler.registry)
```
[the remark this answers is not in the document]
parejkoj: I need this here for ingest to work. Without it I get:

```
lsst.daf.butler.registry._exceptions.DataIdValueError: Could not fetch record for required dimension detector via keys {'instrument': 'DECam', 'detector': 1, 'exposure': 1}.
```

at `self.rawIngestTask.run()`.

python/activator/middleware_interface.py
parejkoj: Ok, this works correctly now; it didn't while I was debugging, but I think I'm finally exporting enough chained collections.

tests/test_middleware_interface.py
parejkoj: No idea. This is from what Eric pushed. Removed.

parejkoj: Ah, it was because of the global `os.environ["IP_APDB"]` in middleware_interface. I've made that a class member now, so it doesn't need to be defined at import.

tests/test_middleware_interface.py
parejkoj: No, it is a requirement for whatever reason. Don't know why.

tests/test_middleware_interface.py
parejkoj: I'm actually building the pipeline now, though: that's how I discovered some of the bugs (like DM-34202). The only thing being mocked is `run`.

python/activator/middleware_interface.py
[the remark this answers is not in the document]
kfindeisen: Ah, because of [this issue](https://github.com/lsst/ap_verify_ci_hits2015/blob/main/scripts/make_preloaded_export.py#L46-L47). Can you clarify the comment to say that, depending on the dataset, the exported data doesn't necessarily include all detectors?

tests/test_middleware_interface.py
kfindeisen: Ok, [that's a problem](https://github.com/lsst-dm/prompt_prototype/blob/main/python/activator/activator.py#L89-L91). I can file a ticket to investigate.

tests/test_middleware_interface.py
kfindeisen: Ok, but does that tell you that you built the right executor, or just any executor? The `where` clause seems like it would be hard to test by any means other than direct introspection.

tests/test_middleware_interface.py
parejkoj: Yes, please file a ticket on this problem, and include some of the middleware folks in the cc:

kfindeisen: Filed as [DM-34241](https://jira.lsstcorp.org/browse/DM-34241).

python/activator/middleware_interface.py
parejkoj: This now has a much better log message.

python/activator/middleware_interface.py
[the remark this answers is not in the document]
parejkoj: My above comment is now out of date, since Jim gave us a fix for the `instrument.register` problem.

parejkoj:
On the team-coding call, we all agreed on not trying to clean up the git history on this across the Eric->John handoff. I cleaned up my own commits as best I could.

PR #22, opened 2022-06-21
DM-35052: Fix single visit restriction for Prompt Processing

kfindeisen:
This is a draft PR, to allow test builds of the activator container.

### DIFF
--- a/pipelines/calibrate.py
+++ b/pipelines/calibrate.py
@@ -1,32 +1,15 @@
 # Config override for lsst.pipe.tasks.calibrate.CalibrateTask
 # This config is for the ap_verify datasets that use gaia/panstarrs as their
 # refcat names.
-from lsst.meas.algorithms import LoadIndexedReferenceObjectsTask
 
 # Use gaia for astrometry (phot_g_mean for everything, as that is the broadest
-# band with the most depth)
-# Use panstarrs for photometry (grizy filters)
-for refObjLoader in (config.astromRefObjLoader,
-                     config.photoRefObjLoader,):
-    refObjLoader.retarget(LoadIndexedReferenceObjectsTask)
+# band with the most depth).
+
 config.connections.astromRefCat = "gaia"
 config.astromRefObjLoader.ref_dataset_name = config.connections.astromRefCat
-config.astromRefObjLoader.filterMap = {
-    "u": "phot_g_mean",
-    "g": "phot_g_mean",
-    "r": "phot_g_mean",
-    "i": "phot_g_mean",
-    "z": "phot_g_mean",
-    "y": "phot_g_mean",
-    "VR": "phot_g_mean"}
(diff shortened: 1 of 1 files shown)


pipelines/calibrate.py
```diff
@@ -1,32 +1,15 @@
...
-config.astromRefObjLoader.filterMap = {
-    "u": "phot_g_mean",
-    "g": "phot_g_mean",
-    "r": "phot_g_mean",
-    "i": "phot_g_mean",
-    "z": "phot_g_mean",
-    "y": "phot_g_mean",
-    "VR": "phot_g_mean"}
+config.astromRefObjLoader.anyFilterMapsToThis = "phot_g_mean"
+config.astromRefObjLoader.filterMap = {}
```
parejkoj, line 11: These two lines are technically not necessary, since [DM-27013](https://jira.lsstcorp.org/browse/DM-27013) is done.

python/activator/middleware_interface.py
```diff
@@ -75,6 +75,12 @@ class MiddlewareInterface:
         appropriate for use in the Google Cloud environment; typically only
         change this when running local tests.
     """
+    _COLL_TEMPLATE = "templates"
```
parejkoj, line 78: Can we call these `_COLLECTION_TEMPLATE` instead, if we're going with an all-caps name?

python/activator/raw.py
```diff
@@ -0,0 +1,47 @@
...
+
+# Format for filenames of raws uploaded to image bucket:
+# instrument/detector/group/snap/expid/filter/*.(fits, fz, fits.gz)
+RAW_REGEXP = re.compile(
+    r"(?P<instrument>.*?)/(?P<detector>\d+)/(?P<group>.*?)/(?P<snap>\d+)/(?P<expid>.*?)/(?P<filter>.*?)/"
+    r"[^/]+\.f"
+)
+
+
+def get_raw_path(instrument, detector, group, snap, exposure_id, filter):
```
parejkoj, line 40: Do we have any way to ensure that this and the regex above stay in sync? A unittest of passing the raw path output into the regex would be good.

parejkoj (approved):
This is a nice improvement. My only real concern is about the "metadata in paths"; a unittest there would help.

Using paths for metadata always makes me nervous, but I guess we have a closed ecosystem here so it should be fine.

pipelines/calibrate.py
kfindeisen: I didn't know that, but for configuration that depends on the input data, I prefer to be explicit instead of trusting that the defaults match up with I really need.

python/activator/raw.py
kfindeisen: Oooh, that's clever!

kfindeisen:
> Using paths for metadata always makes me nervous, but I guess we have a closed ecosystem here so it should be fine.

Note that previously, we were using _the filename_ for metadata, which is even harder to control.

PR #25, opened 2022-09-06
DM-35941: Remove use of implicit dimensions in Prompt Prototype

kfindeisen:
This PR adds explicit `instrument` and `skymap` dimensions to all Butler queries, and removes code that the use of default dimensions had made necessary. The result is much more likely to behave consistently in different conditions, increasing the value and relevance of the existing unit tests. The first few commits fix some smaller bugs discovered while examining the code or testing the explicit dimensions.

This PR depends on, and cannot be merged before, #24.

### DIFF
--- a/python/activator/middleware_interface.py
+++ b/python/activator/middleware_interface.py
@@ -524,6 +524,7 @@ def _query_missing_datasets(src_repo: Butler, dest_repo: Butler,
     *args, **kwargs
         Parameters for describing the dataset query. They have the same
         meanings as the parameters of `lsst.daf.butler.Registry.queryDatasets`.
+        The query must be valid for both ``src_repo`` and ``dest_repo``.
 
     Returns
     -------
--- a/python/activator/middleware_interface.py
+++ b/python/activator/middleware_interface.py
@@ -106,7 +106,6 @@ class MiddlewareInterface:
         define_visits_config = lsst.obs.base.DefineVisitsConfig()
         self.define_visits = lsst.obs.base.DefineVisitsTask(config=define_visits_config, butler=self.butler)
 
-        # TODO DM-34098: note that we currently need to supply instrument here.
         # HACK: explicit collection gets around the fact that we don't have any
         # timestamp/exposure information in a form we can pass to the Butler.
         # This code will break once cameras start being versioned.
@@ -309,6 +308,7 @@ class MiddlewareInterface:
         templates = set(_query_missing_datasets(self.central_butler, self.butler,
                                                 "*Coadd",
                                                 collections=self._COLLECTION_TEMPLATE,
+                                                instrument=self.instrument.getName(),
                                                 skymap=self.skymap_name,
                                                 where=template_where))
         _log.debug("Found %d new template datasets.", len(templates))
@@ -333,6 +333,7 @@ class MiddlewareInterface:
             self.central_butler, self.butler,
             ...,
             collections=self.instrument.makeCalibrationCollectionName(),
+            instrument=self.instrument.getName(),
             where=calib_where))
         if calibs:
             for dataset_type, n_datasets in self._count_by_type(calibs):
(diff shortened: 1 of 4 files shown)


python/activator/middleware_interface.py
```diff
@@ -125,15 +125,18 @@ def __init__(self, central_butler: Butler, image_bucket: str, instrument: str,
...
-        # TODO DM-34098: note that we currently need to supply instrument here.
         # HACK: explicit collection gets around the fact that we don't have any
         # timestamp/exposure information in a form we can pass to the Butler.
         # This code will break once cameras start being versioned.
         self.camera = self.central_butler.get(
             "camera", instrument=self.instrument.getName(),
             collections=self.instrument.makeUnboundedCalibrationRunName()
         )
-        self.skymap = self.central_butler.get("skyMap")
+        # TODO: is central_butler guaranteed to have only one skymap dimension?
```
parejkoj, line 135: Given the assert below, I think you don't need this TODO. We're defining the central repo as having exactly one skymap as part of this, and I think that's something that we as the AP group can ensure.

python/activator/middleware_interface.py
```diff
@@ -165,9 +168,6 @@ def _init_local_butler(self, butler: Butler):
                        ]
         butler.registry.setCollectionChain(self.output_collection, collections)
 
-        # Refresh butler after configuring it, to ensure all required
-        # dimensions and collections are available.
-        butler.registry.refresh()
         self.butler = Butler(butler=butler,
```
parejkoj, line 171: I know it's no longer a refresh, but it might be worth a comment her to say why we make a new butler instance for `self`, instead of using the one that was passed in.

python/activator/middleware_interface.py
```diff
@@ -306,9 +297,8 @@ def _export_refcats(self, export, center, radius):
             Radius to search for refcat shards in.
         """
         indexer = HtmIndexer(depth=7)
-        shards = indexer.getShardIds(center, radius+self.padding)
-        # getShardIds returns a tuple, the first item is the ids list.
-        htm_where = f"htm7 in ({','.join(str(x) for x in shards[0])})"
+        shard_ids, _ = indexer.getShardIds(center, radius+self.padding)
+        htm_where = f"htm7 in ({','.join(str(x) for x in shard_ids)})"
```
parejkoj, line 301: That's a nice cleanup, thanks.

parejkoj (approved):
I'm still slightly grumpy that we can't use constant things like instrument and skymap as butler defaults, but I think I understand why you changed it to this. My grumpiness shouldn't prevent merging; this looks fine otherwise (note a few comment-related comments).

Should we just be storing `instrument.getName()`, or do we need other properties of the `Instrument` object?

python/activator/middleware_interface.py
kfindeisen: I don't see why the `assert` makes the TODO irrelevant. The `assert` is computer-readable but easy for a human to overlook.

I'm not so confident that we will only have one skymap -- I can see a lot of experimentation over which skymap gives e.g. the best performance, especially during commissioning, and there might even be a policy some day that different datasets or data sources are based on different skymaps. Ultimately, it's not our call, but K-T's and the commissioning team's.

kfindeisen:
> I'm still slightly grumpy that we can't use constant things like instrument and skymap as butler defaults, but I think I understand why you changed it to this.

But the problem was that these things _aren't_ constant, especially when the local repo is first being set up -- the code made assumptions that weren't actually true IRL.

> Should we just be storing instrument.getName(), or do we need other properties of the Instrument object?

Well, we use a lot of `Instrument`'s collection-management methods. I'd keep it around just for flexibility, because it's easier to go from an `Instrument` to a name than from a name to an `Instrument`.

timj:
Name to instrument is standard API. We use it all the time for the command-line tooling: `Instrument.fromName()`.

kfindeisen:
Yes, it's proven very useful on this PR. :slightly_smiling_face:

kfindeisen:
@parejkoj, I'm going to try to rearrange the branches so that this one can be merged before `tickets/DM-35053`. Since 9a19f13a79c25b45ccacd0e8e5e44d265891daae and ca496da218fc8343068d819874f0feb5e071d1d5 only make sense after both sets of changes, I'll have to cherry-pick them (including the explanatory comment you requested) to #24 and remove them here.

PR #27, opened 2022-10-05
DM-36080: Separate GCP-specific code in Prompt Processing prototype

kfindeisen:
This PR does a large amount of refactoring centered on `activator.py`, plus some minor reimplementation, with the following goals:

- Removing all LSST dependencies, particularly `Butler` and `Instrument`, from `activator.py`.
- Providing a more platform-independent guarantee that local Butler repos will never collide.
- Separating code within `activator.py` that depends on Google components (Storage and PubSub) from that which does not.
- More unit test coverage of functionality that was previously in `activator.py`, which cannot be imported within the standard `rubin-devl` environment.
- Providing more platform-independent configuration of the databases, especially the APDB.

### DIFF
--- a/python/activator/activator.py
+++ b/python/activator/activator.py
@@ -185,6 +185,7 @@ def next_visit_handler() -> Tuple[str, int]:
             )
             if oid:
                 raw_info = Snap.from_oid(oid)
+                _log.debug("Found %r already present", raw_info)
                 mwi.ingest_image(expected_visit, oid)
                 expid_set.add(raw_info.exp_id)
 
--- a/python/activator/activator.py
+++ b/python/activator/activator.py
@@ -25,7 +25,6 @@ import base64
 import json
 import logging
 import os
-import re
 import time
 from typing import Optional, Tuple
 
@@ -35,7 +34,7 @@ from google.cloud import pubsub_v1, storage
 from .logger import setup_google_logger
 from .make_pgpass import make_pgpass
 from .middleware_interface import get_central_butler, MiddlewareInterface
-from .raw import RAW_REGEXP
+from .raw import Snap
 from .visit import Visit
 
 PROJECT_ID = "prompt-proto"
@@ -185,9 +184,9 @@ def next_visit_handler() -> Tuple[str, int]:
                 expected_visit.detector,
             )
             if oid:
-                m = re.match(RAW_REGEXP, oid)
+                raw_info = Snap.from_oid(oid)
                 mwi.ingest_image(expected_visit, oid)
-                expid_set.add(m.group('expid'))
+                expid_set.add(raw_info.exp_id)
 
         _log.debug(f"Waiting for snaps from {expected_visit}.")
         start = time.time()
@@ -212,20 +211,19 @@ def next_visit_handler() -> Tuple[str, int]:
             for received in response.received_messages:
                 ack_list.append(received.ack_id)
                 oid = received.message.attributes["objectId"]
-                m = re.match(RAW_REGEXP, oid)
-                if m:
-                    instrument, detector, group, snap, expid = m.groups()
-                    _log.debug("instrument, detector, group, snap, expid = %s", m.groups())
+                try:
+                    raw_info = Snap.from_oid(oid)
+                    _log.debug("Received %r", raw_info)
                     if (
-                        instrument == expected_visit.instrument
-                        and int(detector) == int(expected_visit.detector)
-                        and group == str(expected_visit.group)
-                        and int(snap) < int(expected_visit.snaps)
+                        raw_info.instrument == expected_visit.instrument
+                        and raw_info.detector == int(expected_visit.detector)
+                        and raw_info.group == str(expected_visit.group)
+                        and raw_info.snap < int(expected_visit.snaps)
                     ):
                         # Ingest the snap
                         mwi.ingest_image(oid)
-                        expid_set.add(expid)
-                else:
+                        expid_set.add(raw_info.exp_id)
+                except ValueError:
                     _log.error(f"Failed to match object id '{oid}'")
             subscriber.acknowledge(subscription=subscription.name, ack_ids=ack_list)
 
--- a/python/activator/activator.py
+++ b/python/activator/activator.py
@@ -44,14 +44,14 @@ PROJECT_ID = "prompt-proto"
 verification_token = os.environ["PUBSUB_VERIFICATION_TOKEN"]
 # The full instrument class name, including module path.
 config_instrument = os.environ["RUBIN_INSTRUMENT"]
-active_instrument = Instrument.from_string(config_instrument)
+instrument_name = Instrument.from_string(config_instrument).getName()
 calib_repo = os.environ["CALIB_REPO"]
 image_bucket = os.environ["IMAGE_BUCKET"]
 timeout = os.environ.get("IMAGE_TIMEOUT", 50)
 local_repos = os.environ.get("LOCAL_REPOS", "/tmp")
 
 setup_google_logger(
-    labels={"instrument": active_instrument.getName()},
+    labels={"instrument": instrument_name},
 )
 _log = logging.getLogger("lsst." + __name__)
 _log.setLevel(logging.DEBUG)
@@ -67,7 +67,7 @@ app = Flask(__name__)
 subscriber = pubsub_v1.SubscriberClient()
 topic_path = subscriber.topic_path(
     PROJECT_ID,
-    f"{active_instrument.getName()}-image",
+    f"{instrument_name}-image",
 )
 subscription = None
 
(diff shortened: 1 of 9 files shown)


python/activator/activator.py
```diff
@@ -25,40 +25,37 @@
 import json
 import logging
 import os
-import re
 import time
 from typing import Optional, Tuple
 
 from flask import Flask, request
 from google.cloud import pubsub_v1, storage
```
dspeck1, line 32: Remove for now and replace with boto3 and confluent kafka packages?

python/activator/activator.py
```diff
@@ -71,25 +68,17 @@
 subscriber = pubsub_v1.SubscriberClient()
 topic_path = subscriber.topic_path(
     PROJECT_ID,
-    f"{active_instrument.getName()}-image",
+    f"{instrument_name}-image",
 )
 subscription = None
 
 storage_client = storage.Client()
```
dspeck1, line 75: Remove for GCS dependency

dspeck1:
Added minor comments on Google Cloud Storage imports and storage client library still being there.  Fine for now, but if you were intending to remove all GCP dependencies.

kfindeisen:
> Added minor comments on Google Cloud Storage imports and storage client library still being there. Fine for now, but if you were intending to remove all GCP dependencies.

No, the goal was to remove everything that _wasn't_ a GCP dependency... at least, as far as practical.

python/activator/middleware_interface.py
```diff
@@ -45,6 +45,35 @@
 _log.setLevel(logging.DEBUG)
 
 
+def get_central_butler(central_repo: str, instrument_class: str):
+    """Provide a useful Butler that can access the given repository and work on
+    data for the given instrument.
```
parejkoj, line 50: "useful Butler"?

python/activator/middleware_interface.py
```diff
@@ -45,6 +45,35 @@
...
+        A Butler for ``central_repo`` pre-configured to load
+        ``instrument_name`` data.
+    """
+    # TODO: how much overhead do we take on from asking the central repo about
+    # the instrument instead of handling it internally?
+    registry = Butler(central_repo).registry
+    instrument = lsst.obs.base.Instrument.from_string(instrument_class, registry)
+    return Butler(central_repo,
+                  collections=[instrument.makeCollectionName("defaults")],
+                  writeable=True,
```
parejkoj, line 72: Why does this one need to be writeable; isn't this just for accessing calibs and templates? If it really does, that should probably be noted in the Returns comment.

python/activator/middleware_interface.py
```diff
@@ -141,18 +174,32 @@ def __init__(self, central_butler: Butler, image_bucket: str, instrument: str,
         # How much to pad the refcat region we will copy over.
         self.padding = 30*lsst.geom.arcseconds
 
-    def _init_local_butler(self, butler: Butler):
+    def _make_apdb_uri(self):
+        """Generate a URI for accessing the APDB.
+        """
+        # TODO: merge this code with make_pgpass.py
```
parejkoj, line 180: Is this something that could live in `analysis_ap`? There's code for attaching to an APDB in there, too.

python/activator/raw.py
```diff
@@ -25,10 +25,64 @@
...
+    detector: int
+    group: str         # observatory-specific ID; not the same as visit number
+    snap: int          # exposure number within a group
+    exp_id: int        # Butler-compatible unique exposure ID
+    filter: str        # physical filter
+
+    def __str__(self):
+        """Return a short string that disambiguates the image.
+        """
+        return f"(exposure {self.exp_id}, group {self.group}/{self.snap})"
```
parejkoj, line 48: Do we want an `=` after the names, instead of spaces?

tests/test_middleware_interface.py
```diff
@@ -144,7 +147,19 @@ def setUp(self):
...
         # TemporaryDirectory warns on leaks
-        self.repo.cleanup()
+        self.interface._repo.cleanup()  # TODO: should MiddlewareInterface have a cleanup method?
+        self.workspace.cleanup()
+
+    def test_get_butler(self):
+        for butler in [get_central_butler(self.central_repo, "lsst.obs.decam.DarkEnergyCamera"),
+                       get_central_butler(self.central_repo, instname),
+                       ]:
+            # TODO: better way to test repo location?
```
parejkoj, line 157: Unfortunately, I don't have any ideas here.

tests/test_raw.py
```diff
@@ -55,3 +67,47 @@ def test_writeread(self):
...
+        self.assertEqual(parsed.detector, self.detector)
+        self.assertEqual(parsed.group, self.group)
+        self.assertEqual(parsed.snap, self.snap)
+        self.assertEqual(parsed.exp_id, self.exposure)
+        self.assertEqual(parsed.filter, self.filter)
+
+    def test_bad_snap(self):
+        path = get_raw_path(self.instrument, f"{self.detector}b", self.group,
+                            self.snap, self.exposure, self.filter)
+        with self.assertRaises(ValueError):
```
parejkoj, line 88: Worth a `assertRaisesRegex` to ensure that you're getting the ValueError you expected?

parejkoj (commented):
Looks good in general: a few comments on labelling things.

There are several new TODOs here: should they get tickets? We probably do want a "figure out how much overhead various steps cause" ticket in the future, which is what some of the TODOs are for. If they're just on instantiating the class, it may not be a problem, unless we instantiate a new class for each Visit.

python/activator/activator.py
```diff
@@ -123,6 +112,35 @@ def check_for_snap(
...
+    """
+    envelope = http_request.get_json()
+    if not envelope:
+        raise ValueError("no Pub/Sub message received")
+    if not isinstance(envelope, dict) or "message" not in envelope:
+        raise ValueError("invalid Pub/Sub message format")
+
+    payload = base64.b64decode(envelope["message"]["data"])
+    data = json.loads(payload)
+    return Visit(**data)
```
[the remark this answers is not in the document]
kfindeisen: No, `Visit` accepts the input as-is (see [my question on `#dm-prompt-processing`](https://lsstc.slack.com/archives/C0345PUU4N9/p1661454637158869)). But that's a pre-existing issue, this just factored that code out of the handler.

python/activator/activator.py
```diff
@@ -175,9 +184,10 @@ def next_visit_handler() -> Tuple[str, int]:
                 expected_visit.detector,
             )
             if oid:
-                m = re.match(RAW_REGEXP, oid)
+                raw_info = Snap.from_oid(oid)
+                _log.debug("Found %r already arrived", raw_info)
```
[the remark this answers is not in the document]
kfindeisen: The raw was already in the bucket before the handler started listening for arrival messages.

python/activator/logger.py
```diff
@@ -19,12 +19,38 @@
...
 
-__all__ = ["GCloudStructuredLogFormatter"]
+__all__ = ["GCloudStructuredLogFormatter", "setup_google_logger"]
 
 import json
 import logging
 
 
+# TODO: replace with something more extensible, once we know what needs to
+# vary besides the formatter (handler type?).
```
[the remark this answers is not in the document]
kfindeisen: This is an architectural issue -- having a hardcoded function for each possible environment just feels wrong, and it's not clear to me how to design it better. If it turns out that we don't need anything that extensible after the USDF port is done (or especially after we start field testing with AuxTel), then we can just remove the TODO.

python/activator/middleware_interface.py
kfindeisen: Good catch; I'd written the docs before DM-35053 merged and forgot to update them.

python/activator/middleware_interface.py
kfindeisen: This is  just parsing the PP service's configuration (i.e., its envvars); what common ground would it have with `analysis_ap`? In any case, I don't think it's a good idea for the pipeline to depend on analysis packages.

tests/test_raw.py
kfindeisen: I don't see why the error message matters; what's important is that the non-matching path is detected and can be handled.

python/activator/activator.py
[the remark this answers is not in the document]
parejkoj: Hmm, I'm not sure how else I'd word that. Maybe it makes sense in the logging context?

tests/test_raw.py
parejkoj: There may be other things that raise `ValueError`, so I like to ensure that the error was raised at the place I expected it. Easiest way to check that is via the message.

python/activator/activator.py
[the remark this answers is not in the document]
kfindeisen: Would "already present" or "already in the bucket" be clearer?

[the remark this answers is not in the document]
parejkoj: I think either of those are clearer, thank you.

PR #33, opened 2022-10-19
DM-36602: Move from NCSA+GCP to USDF

hsinfang:


bin/prompt_prototype_upload_raws.sh
```diff
@@ -20,37 +20,47 @@
...
 # instrument/detector/group/snap/exposureId/filter/instrument-group-snap-exposureId-filter-detector
-gsutil cp "${RAW_DIR}/HSCA05913553.fits" \
-    gs://${UPLOAD_BUCKET}/HSC/0/2016030700001/0/0059134/HSC-G/HSC-2016030700001-0-0059134-HSC-G-0.fits
-gsutil cp "${RAW_DIR}/HSCA05913542.fits" \
-    gs://${UPLOAD_BUCKET}/HSC/4/2016030700001/0/0059134/HSC-G/HSC-2016030700001-0-0059134-HSC-G-4.fits
-gsutil cp "${RAW_DIR}/HSCA05913543.fits" \
-    gs://${UPLOAD_BUCKET}/HSC/5/2016030700001/0/0059134/HSC-G/HSC-2016030700001-0-0059134-HSC-G-5.fits
+aws --endpoint-url https://s3dfrgw.slac.stanford.edu s3 \
+    cp "${RAW_DIR}/HSCA05913553.fits" \
+    s3://${UPLOAD_BUCKET}/HSC/0/2016030700001/0/0059134/HSC-G/HSC-2016030700001-0-0059134-HSC-G-0.fits
```
kfindeisen, line 36: I don't have an `aws` alias(?) defined. Would it be possible to provide a definition local to this script, so that it runs equally well for all Rubin users?

bin/prompt_prototype_upload_raws.sh
```diff
@@ -20,37 +20,47 @@
...
+# This script uploads the raw files from the HSC PDR2 run to a bucket at USDF.
+# It renames the files to match prompt_prototype conventions. The user must
+# have bucket access already configured.
 
 set -e  # Abort on any error
 
-RAW_DIR="/datasets/hsc/raw/ssp_pdr2/2016-03-07"
-UPLOAD_BUCKET=rubin-prompt-proto-unobserved
+RAW_DIR="/sdf/group/rubin/datasets/hsc/raw/ssp_pdr2/2016-03-07"
+UPLOAD_BUCKET=rubin-pp-users/unobserved
```
kfindeisen, line 30: Would it be possible to update the Buckets / USDF section of the Playbook to mention that `rubin-pp-users` is used for the unobserved files? There's analogous language in the Google Cloud section.

bin/prompt_prototype_upload_raws.sh
```diff
@@ -20,37 +20,47 @@
...
 
 # Filename format is defined in activator/raw.py:
 # instrument/detector/group/snap/exposureId/filter/instrument-group-snap-exposureId-filter-detector
-gsutil cp "${RAW_DIR}/HSCA05913553.fits" \
-    gs://${UPLOAD_BUCKET}/HSC/0/2016030700001/0/0059134/HSC-G/HSC-2016030700001-0-0059134-HSC-G-0.fits
-gsutil cp "${RAW_DIR}/HSCA05913542.fits" \
-    gs://${UPLOAD_BUCKET}/HSC/4/2016030700001/0/0059134/HSC-G/HSC-2016030700001-0-0059134-HSC-G-4.fits
-gsutil cp "${RAW_DIR}/HSCA05913543.fits" \
-    gs://${UPLOAD_BUCKET}/HSC/5/2016030700001/0/0059134/HSC-G/HSC-2016030700001-0-0059134-HSC-G-5.fits
+aws --endpoint-url https://s3dfrgw.slac.stanford.edu s3 \
```
kfindeisen, line 34: Perhaps:
```suggestion
aws --endpoint-url ${S3_ENDPOINT_URL} s3 \
```
combined with a local definition of the variable (per my comment about `aws`)?

kfindeisen (commented):
Many thanks for taking this on! However, I'm a bit concerned about this script depending on how the user's aliases are configured; clarification on this point would be appreciated.

kfindeisen (approved):
Thanks for the changes! It looks good.

bin/prompt_prototype_upload_raws.sh
hsinfang: I actually had installed `aws` command line interface locally, but using `singularity` does seem nicer. I have added an alias in the script so users can use it without needing to install/setup extra things.

PR #34, opened 2022-10-26
DM-36712: Migrate activator.py to use Kafka/Knative

kfindeisen:
This PR replaces all Google-specific components (including the logger, which doesn't depend on Google APIs as such) with USDF-appropriate components, and documents how to run the activator on our Kubernetes cluster.

### DIFF
--- a/doc/playbook.rst
+++ b/doc/playbook.rst
@@ -130,6 +130,9 @@ The buckets can be inspected by calling the following from ``rubin-devl``:
 Prototype Service
 =================
 
+Google Cloud
+------------
+
 The service can be controlled by Google Cloud Run, which will automatically trigger instances based on ``nextVisit`` messages and can autoscale the number of them depending on load.
 Each time the service container is updated, a new revision of the service should be edited and deployed.
 (Continuous deployment has not yet been set up.)
@@ -192,6 +195,56 @@ It accepts messages and gateways them to Cloud Run.
 * Set the acknowledgement deadline to 600 seconds
 * Set the "Retry policy" to "Retry immediately"
 
+USDF
+----
+
+The service can be controlled with ``kubectl`` from ``rubin-devl``.
+You must first `get credentials for the development cluster <https://k8s.slac.stanford.edu/usdf-prompt-processing-dev>`_ on the web; ignore the installation instructions and copy the commands from the second box.
+Credentials are good for roughly one work day.
+
+Each time the service container is updated, a new revision of the service should be edited and deployed.
+(Continuous deployment has not yet been set up.)
+To create the service, clone the `slaclab/rubin-usdf-prompt-processing`_ repo and navigate to the ``kubernetes/overlays/dev/prompt-proto-service`` directory.
+Edit ``prompt-proto-service.yaml`` to point to the new service container (likely a ticket branch instead of ``latest``), then run ``make apply`` *from the same directory*.
+See the readme in that directory for more details.
+
+.. _slaclab/rubin-usdf-prompt-processing: https://github.com/slaclab/rubin-usdf-prompt-processing/
+
+All service configuration is in ``prompt-proto-service.yaml``.
+It includes the following required environment variables:
+
+* RUBIN_INSTRUMENT: the "short" instrument name
+* PUBSUB_VERIFICATION_TOKEN: choose an arbitrary string matching the Pub/Sub endpoint URL below.
+  This variable is currently unused and may be removed in the future.
+* IMAGE_BUCKET: bucket containing raw images
+* CALIB_REPO: URI to repo containing calibrations (and templates)
+* IP_APDB: IP address or hostname and port of the APDB (see `Databases`_, below)
+* IP_REGISTRY: IP address or hostname and port of the registry database (see `Databases`_)
+* DB_APDB: PostgreSQL database name for the APDB
+* PSQL_APDB_PASS: secret containing the password for ``USER_APDB`` (see below)
+* DB_REGISTRY: PostgreSQL database name for the registry database
+* PSQL_REGISTRY_PASS: secret containing the password for ``USER_REGISTRY`` (see below)
+* KAFKA_CLUSTER: hostname and port of the Kafka provider
+
+The following environment variables are optional:
+
+* IMAGE_TIMEOUT: timeout in seconds to wait for raw image, default 50 sec.
+* LOCAL_REPOS: absolute path (in the container) where local repos are created, default ``/tmp``.
+* USER_APDB: database user for the APDB, default "postgres"
+* USER_REGISTRY: database user for the registry database, default "postgres"
+* NAMESPACE_APDB: the database namespace for the APDB, defaults to the DB's default namespace
+
+Secrets are configured through the makefile and ``kustomization.yaml``.
+
+A few useful commands for managing the service:
+
+* ``kubectl config set-context usdf-prompt-processing-dev --namespace=prompt-proto-service`` sets the default namespace for the following ``kubectl`` commands to ``prompt-proto-service``.
+  Note that many of the workflows in `slaclab/rubin-usdf-prompt-processing`_ run in the ``knative-serving`` or ``knative-eventing`` namespaces; to examine the resources of these workflows, add e.g. ``-n knative-eventing`` to the examples below.
+* ``kubectl get serving`` summarizes the state of the service, including which revision(s) are currently handling messages.
+  A revision with 0 replicas is inactive.
+* ``kubectl get pods`` lists the Kubernetes pods that are currently running, how long they have been active, and how recently they crashed.
+* ``kubectl logs <pod>`` outputs the entire log associated with a particular pod.
+  This can be a long file, so consider piping to ``less`` or ``grep``.
 
 tester
 ======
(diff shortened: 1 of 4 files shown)


doc/playbook.rst
```diff
@@ -192,6 +195,54 @@ It accepts messages and gateways them to Cloud Run.
...
+* LOCAL_REPOS: absolute path (in the container) where local repos are created, default ``/tmp``.
+* USER_APDB: database user for the APDB, default "postgres"
+* USER_REGISTRY: database user for the registry database, default "postgres"
+* NAMESPACE_APDB: the database namespace for the APDB, defaults to the DB's default namespace
+
+Secrets are configured through the makefile and ``kustomization.yaml``.
+
+A few useful commands for managing the service:
+
+* ``kubectl get serving -n prompt-proto-service`` summarizes the state of the service, including which revision(s) are currently handling messages.
```
ktlim, line 241: If you say `kubectl config set-context usdf-prompt-processing-dev --namespace=prompt-proto-service` I believe you can eliminate the `-n` option from all of these command lines.

python/activator/activator.py
```diff
@@ -50,8 +51,15 @@
 timeout = os.environ.get("IMAGE_TIMEOUT", 50)
 # Absolute path on this worker's system where local repos may be created
 local_repos = os.environ.get("LOCAL_REPOS", "/tmp")
-
-setup_google_logger(
+# Kafka server
+kafka_cluster = os.environ["KAFKA_CLUSTER"]
+# Kafka group; must be worker-unique to keep workers from "stealing" messages for others.
+kafka_group_id = str(uuid.uuid4())
```
ktlim, line 57: Why `uuid4` and not `uuid1`?

python/activator/activator.py
```diff
@@ -50,8 +51,15 @@
...
+# The topic on which to listen to updates to image_bucket
+# bucket_topic = f"{instrument_name}-image"
+bucket_topic = "pp-bucket-notify-topic"
```
ktlim, line 60: Should this be configurable (i.e. via env var)?

python/activator/activator.py
```diff
@@ -65,14 +73,13 @@
...
-subscriber = pubsub_v1.SubscriberClient()
-topic_path = subscriber.topic_path(
-    PROJECT_ID,
-    f"{instrument_name}-image",
-)
-subscription = None
+consumer = kafka.Consumer({
+    "bootstrap.servers": kafka_cluster,
+    "group.id": kafka_group_id,
+    "auto.offset.reset": "earliest",
```
ktlim, line 79: `earliest` may be overkill.  The desired message for this worker can't show up until after the next_visit event that triggered the worker.  But if cold start takes a while, I guess it's possible that the image arrived during that time, especially for images with shorter exposure times.

Note that the dynamic subscription I was using before for Pub/Sub is the equivalent of `latest`, not `earliest`.

We'll need to expire messages from the Kafka queues relatively rapidly if we don't want to have to search through a lot of old images.  I believe the default is 7 days.

python/activator/activator.py
```diff
@@ -155,13 +214,8 @@ def next_visit_handler() -> Tuple[str, int]:
...
         The HTTP response status code to return to the client.
     """
-    if request.args.get("token", "") != verification_token:
-        return "Invalid request", 400
-    subscription = subscriber.create_subscription(
-        topic=topic_path,
-        ack_deadline_seconds=60,
-    )
-    _log.debug(f"Created subscription '{subscription.name}'")
+    consumer.subscribe([bucket_topic])
```
ktlim, line 217: Because there's no explicit redelivery on unprocessed messages like Pub/Sub had, there's no need to add subscriptions dynamically, I think.  Can't we subscribe once and reuse for each visit?

python/activator/activator.py
```diff
@@ -192,13 +246,13 @@ def next_visit_handler() -> Tuple[str, int]:
...
+            response = consumer.consume(
+                num_messages=189 + 8 + 8,
                 timeout=timeout,
             )
             end = time.time()
-            if len(response.received_messages) == 0:
+            messages = _filter_messages(response)
+            if len(messages) == 0:
                 if end - start < timeout:
                     _log.debug(f"Empty pull after {end - start}s for {expected_visit}.")
```
ktlim, line 257: Maybe change to `consume` from `pull`?

python/activator/activator.py
```diff
@@ -208,20 +262,19 @@ def next_visit_handler() -> Tuple[str, int]:
...
+                    try:
+                        raw_info = Snap.from_oid(oid)
+                        _log.debug("Received %r", raw_info)
+                        if raw_info.is_consistent(expected_visit):
+                            # Ingest the snap
+                            mwi.ingest_image(oid)
+                            expid_set.add(raw_info.exp_id)
+                    except ValueError:
+                        _log.error(f"Failed to match object id '{oid}'")
+            # Don't commit message, it causes other messages to be dropped as well.
```
ktlim, line 276: Which other messages get dropped?  At worst, I would think we can commit when we've processed everything that was returned by `consume()`.

ktlim (approved):

doc/playbook.rst
```diff
@@ -192,6 +195,54 @@ It accepts messages and gateways them to Cloud Run.
 * Set the acknowledgement deadline to 600 seconds
 * Set the "Retry policy" to "Retry immediately"
 
+USDF
+----
+
+The service can be controlled with ``kubectl`` from ``rubin-devl``.
+You must first `get credentials for the development cluster <https://k8s.slac.stanford.edu/usdf-prompt-processing-dev>`_ on the web; ignore the installation instructions and copy the commands from the second box.
+Credentials are good for roughly one work day.
```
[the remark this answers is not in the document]
kfindeisen: IIRC the figure I was given for Unix S3DF was 8 hours. But I'll admit I hadn't tried _not_ getting a new token at the start of the day.

python/activator/activator.py
kfindeisen: Because the [uuid docs](https://docs.python.org/3/library/uuid.html) warned that `uuid1` has weaker concurrency guarantees, and that would make a difference if e.g. we ever have multiple workers per container. I really want to avoid the scenario we had on GCP where we tried to disambiguate on the PID, but every container always started a worker with PID 503.

Is there a reason `uuid1` would be preferable to `uuid4`?

python/activator/activator.py
kfindeisen: I'm worried about how well that would work once we have topics that depend on the instrument, detector, etc.

python/activator/activator.py
timj: Butler uses uuid4.

python/activator/activator.py
kfindeisen: This was based on what I'd read in the [Kafka docs](https://docs.confluent.io/platform/current/clients/consumer.html) and [a comment on Slack](https://lsstc.slack.com/archives/C0345PUU4N9/p1666802736907339?thread_ts=1666740764.692819&cid=C0345PUU4N9) from @dspeck1. Maybe I misunderstood, but I thought Kafka can only commit ranges of messages, and that potentially meant messages for other detectors or (with enough mixing) even visits.

python/activator/activator.py
kfindeisen: This was cargo-culted from [test-consumer](https://github.com/slaclab/rubin-usdf-prompt-processing/blob/main/test-consumer/main.py). (Sorry!)

I think this might be tied to your question below about `subscribe` -- can we actually make assumptions about the relative ordering of messages (`next_visit` vs. bucket notification) transmitted over different channels?

python/activator/activator.py
ktlim: If we had a worker per IP, then uuid1 might give some debuggability.  But that's probably available elsewhere.  No big deal.

python/activator/activator.py
ktlim: Instrument has to have a different set of workers, I would think.  You wouldn't be dynamically switching a given worker to a different instrument.

Detector/raft might be a problem.  At least the non-dynamic part should be configurable, then?  Or have the configuration specify a template?

python/activator/activator.py
ktlim: The point to the Slack message was that a given _consumer group_ commits all at once.  Since every worker will have its own consumer group, it's not a problem to commit everything it has seen.

python/activator/activator.py
ktlim: I think the worst case scenario is biases, where each exposure might be ~2 seconds in length (as that is the readout time).  next_visit would then be a minimum of 4 seconds before readout finishes (sent prior to the previous bias) and at least 4.5 seconds before the file would be compressed and copied from the CCS (at the Summit; likely more at USDF).

The normal science case with single snaps is at least 64.5 seconds before the first file could arrive; 49.5 seconds for two snaps.

Having the next_visit message fall so far behind that the image notification arrives first thus seems highly unlikely.  But I'm OK with `earliest` if we set the Kafka queues to no more than a day or even an hour of retention.

python/activator/activator.py
ktlim: I guess I'm mostly thinking of a deployment per instrument and so having the instrument be part of the configuration and not the code is important.

kfindeisen: The instrument name itself is already part of the configuration -- we need it for lots of things!

PR #35, opened 2022-11-04
DM-36720: migrate upload.py to USDF

hsinfang:


python/tester/upload.py
```diff
@@ -94,12 +95,12 @@ def send_next_visit(publisher, group, visit_infos):
         represent multiple snaps.
     """
     _log.info(f"Sending next_visit for group: {group}")
-    topic_path = publisher.topic_path(PROJECT_ID, "nextVisit")
+    topic = "nextVisit"
```
kfindeisen, line 98: Is this still the correct topic name? The [`overlays/dev`](https://github.com/slaclab/rubin-usdf-prompt-processing/blob/main/kubernetes/overlays/dev/kafka/next-visit-topic.yaml) uses `"next-visit-topic"`, and we may want separate topics for the real-world and development environments.

kfindeisen (approved):
Thanks for doing the conversion work, and for adding unit tests for the uploader! I'm looking forward to seeing all the pieces fit together.

python/tester/upload.py
```diff
@@ -1,11 +1,15 @@
+__all__ = ["get_last_group", ]
+
+import boto3
```
timj, line 3: What functionality is missing from lsst.resources that would make it work for this? Flexibility of endpoint (GCS vs S3 vs WebDAV) is what lsst.resources was designed to help with so it would help me if I knew why you had decided against using it. Thanks.

kfindeisen: This particular script is not supposed to depend on anything LSST. It was previously only useful in an environment that didn't have the Stack installed, and may be so again in the future.

timj: But you are installing boto3 so you are installing external software. Why can't you install lsst-resources from PyPI as well? Or are you using conda-forge so want me to add lsst-resources to conda-forge instead? lsst-resources is completely standalone PyPI installable with BSD license.

python/tester/upload.py
hsinfang: Or should it be `pp-bucket-notify-topic` from the [activator|https://github.com/lsst-dm/prompt_prototype/blob/main/python/activator/activator.py#L60] if I read that correctly? 
Yeah we probably want to decide what topics to use for real-world and dev for all components?

kfindeisen: I agree that the same question applies to `pp-bucket-notify-topic`, but we probably shouldn't use it for `next_visit` in addition to image arrival notifications.

python/tester/upload.py
```diff
@@ -138,18 +141,19 @@ def main():
     credentials = service_account.Credentials.from_service_account_file(
         "./prompt-proto-upload.json"
     )
-    storage_client = storage.Client(PROJECT_ID, credentials=credentials)
-    dest_bucket = storage_client.bucket("rubin-prompt-proto-main")
+    endpoint_url = "https://s3dfrgw.slac.stanford.edu"
```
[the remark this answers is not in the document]
kfindeisen: That's surprising, since `activator.py` didn't need to (though it was using `client` instead of `resource`).

python/tester/upload.py
timj: `lsst.resources` is how butler datastore can work on S3, WebDAV, GCS, or local file system without butler having to care where the files end up.

python/tester/upload.py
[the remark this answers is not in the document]
hsinfang: To be clear, I can instantiate a `client` without an explicit endpoint, but will then fail to access anything as the endpoint seems wrong: 
`botocore.exceptions.ClientError: An error occurred (InvalidAccessKeyId) when calling the ListObjects operation: The AWS Access Key Id you provided does not exist in our records.`

tests/test_upload.py
```diff
@@ -0,0 +1,79 @@
...
+        s3 = boto3.resource("s3")
+        bucket = s3.Bucket(self.bucket_name)
+        try:
+            bucket.objects.all().delete()
+        except botocore.exceptions.ClientError as e:
+            if e.response["Error"]["Code"] == "404":
+                # the key was not reachable - pass
+                pass
+            else:
+                raise
```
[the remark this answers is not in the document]
hsinfang: I cargo-cult-ed it from [daf_butler](https://github.com/lsst/daf_butler/blob/main/tests/test_butler.py#L1845-L1855) and it's in [lsst.resources](https://github.com/lsst/resources/blob/main/tests/test_s3.py#L58) too. 

The `moto` [docs](https://docs.getmoto.org/en/latest/docs/getting_started.html#unittest-usage) looks like it might not be necessary.

tests/test_upload.py
```diff
@@ -0,0 +1,79 @@
...
+from moto import mock_s3
+
+from activator.raw import get_raw_path
+from tester.upload import get_last_group
+
+
+class UploadTest(unittest.TestCase):
+    """Test components in the upload tester.
+    """
+    mock_s3 = mock_s3()
```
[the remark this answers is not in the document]
hsinfang: Seems to be how it's intended to be used from its [doc](https://docs.getmoto.org/en/latest/docs/getting_started.html#unittest-usage).

python/tester/upload.py
```diff
@@ -175,8 +175,8 @@ def get_last_group(storage_client, instrument, date):
     Returns
     -------
     group : `int`
-        The largest existing group for ``instrument``, or a newly generated
-        group if none exist.
+        The largest existing group for ``instrument`` plus a random number,
+        or a newly generated group if none exist.
```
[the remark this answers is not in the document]
hsinfang: In practice as the number is incremented before passing to the functions ( https://github.com/lsst-dm/prompt_prototype/blob/main/python/tester/upload.py#L157-L160 ), the addition of a random number here is not really necessary.  Mind if I remove it from here, so this really "gets last group" (instead of the last group plus something)? I don't think any other code uses this `upload.py`, right? 

If we want to keep some randomness, I can add a random number in `main` (for passing into `upload_from_raws` and `upload_from_random`).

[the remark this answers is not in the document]
hsinfang: https://github.com/lsst-dm/prompt_prototype/pull/35/commits/08131a880d48787ba88282f4f22baff6586addb5

tests/test_upload.py
```diff
@@ -0,0 +1,79 @@
...
+        s3 = boto3.resource("s3")
+        bucket = s3.Bucket(self.bucket_name)
+
+        random_call.return_value = 4
+        last_group = get_last_group(bucket, "TestCam", "20221102")
+        self.assertEqual(last_group, 2022110200002 + random_call.return_value)
+
+        # Test the case of no match
+        last_group = get_last_group(bucket, "TestCam", "20110101")
+        self.assertEqual(last_group, int(20110101) * 100_000)
```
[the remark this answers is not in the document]
hsinfang: The randomness in `get_last_group` has been moved out, so now it really gives the last group number.

python/tester/upload.py
hsinfang: I'm changing this to `next-visit-topic` for now. When we find out what name Summit really uses it can be changed easily.

PR #36, opened 2022-11-14
DM-36995: Remove test_upload.py temporarily

hsinfang:


kfindeisen (approved):
Looks good; I used https://github.com/lsst-dm/prompt_prototype/actions/runs/3465053649 to check that the tests also pass in GHA.

PR #38, opened 2022-11-15
DM-36999: Complete integration test at USDF

kfindeisen:
This PR contains a number of miscellaneous fixes that prevented an end-to-end run on USDF. Many of these are preexisting bugs that were never exercised on GCP.

### DIFF
--- a/python/activator/middleware_interface.py
+++ b/python/activator/middleware_interface.py
@@ -136,6 +136,8 @@ class MiddlewareInterface:
     #   pre-made inputs, and raws, in that order. However, self.butler is not
     #   guaranteed to contain concrete data, or even the dimensions
     #   corresponding to self.camera and self.skymap.
+    # if it exists, the latest run in self.output_collection is always the first
+    # in the chain.
 
     def __init__(self, central_butler: Butler, image_bucket: str, instrument: str,
                  local_storage: str,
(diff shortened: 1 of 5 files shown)


python/activator/middleware_interface.py
```diff
@@ -663,16 +665,21 @@ def export_outputs(self, visit: Visit, exposure_ids: set[int]) -> None:
...
-        self._export_subset(visit, exposure_ids,
-                            # TODO: find a way to merge datasets like *_config
-                            # or *_schema that are duplicated across multiple
-                            # workers.
-                            self._get_safe_dataset_types(self.butler),
-                            self.output_collection + "/*",  # exclude inputs
-                            umbrella)
-        _log.info(f"Pipeline products saved to collection '{umbrella}' for "
-                  f"detector {visit.detector} of {exposure_ids}.")
+        latest_run = self.butler.registry.getCollectionChain(self.output_collection)[0]
```
hsinfang, line 670: Is this an implementation detail that may get changed in the Rubin middleware? Would it make sense to ask Middleware to implement a public method that gives the latest run?

kfindeisen: No, `MiddlewareInterface` manages the collections purely internally, which is why I made the run order a class invariant. At present, the invariant is enforced by only adding runs to the chain by calling `_prepend_collection`.

hsinfang: Never mind... My questions is answered by [this line ](https://github.com/lsst-dm/prompt_prototype/blob/9b4a6e715b1c3db3868390a6108abef8b3e5883b/python/activator/middleware_interface.py#L135-L136)

tests/test_middleware_interface.py
```diff
@@ -567,3 +580,27 @@ def test_export_outputs_bad_visit(self):
...
+        self._simulate_run()
+        self.interface.export_outputs(self.second_visit, {self.raw_data_id["exposure"]})
+
+        central_butler = Butler(self.central_repo.name, writeable=False)
+        raw_collection = f"{instname}/raw/all"
+        export_collection = f"{instname}/prompt-results"
+        self._check_datasets(central_butler,
+                             "raw", raw_collection, 2, self.raw_data_id)
+        # Did not export raws directly to raw/all.
+        self.assertNotEqual(central_butler.registry.getCollectionType(raw_collection), CollectionType.RUN)
```
hsinfang, line 597: Do we expect this collection to be a chained collection? Why not `assertEqual` that instead?

python/tester/upload.py
```diff
@@ -147,19 +147,23 @@ def main():
...
+        new_group_base = last_group + random.randrange(10, 19)
+        if raw_pool:
+            _log.info(f"Observing real raw files from {instrument}.")
+            upload_from_raws(producer, instrument, raw_pool, src_bucket, dest_bucket,
+                             n_groups, new_group_base)
+        else:
+            _log.info(f"No raw files found for {instrument}, generating dummy files instead.")
+            upload_from_random(producer, instrument, dest_bucket, n_groups, new_group_base)
+    finally:
+        producer.flush(30.0)
```
hsinfang, line 166: I see the warning but I'm not sure if it matters to flush and wait here. Don't they get sent out still?

hsinfang (approved):

python/tester/upload.py
kfindeisen: Yes, the messages get sent, but such warnings usually mean that a class finalizer is doing cleanup that it can't be trusted with. Also, if somebody later adds callbacks to the message code, they will only work correctly if we flush.

As for waiting, the delay when I tested it was too small to measure.

PR #41, opened 2022-12-02
DM-37068: Clear out conflicting datasets before prompt processing runs

kfindeisen:
This PR adds code for removing shared (usually init-output) datasets from an output run, making it possible to reuse runs for multiple groups and detectors. Actually implementing reuse is deferred to [DM-36586](https://jira.lsstcorp.org/browse/DM-36586).

The deletion is a stopgap until [DM-36162](https://jira.lsstcorp.org/browse/DM-36162) gives us better tools for handling preexisting or conflicting datasets, so the new code is all written on the assumption that it will be removed later.

### DIFF
--- a/tests/test_middleware_interface.py
+++ b/tests/test_middleware_interface.py
@@ -31,7 +31,9 @@ import astropy.coordinates
 import astropy.units as u
 
 import astro_metadata_translator
+import lsst.pex.config
 import lsst.afw.image
+import lsst.afw.table
 from lsst.daf.butler import Butler, CollectionType, DataCoordinate
 import lsst.daf.butler.tests as butler_tests
 from lsst.obs.base.formatters.fitsExposure import FitsImageFormatter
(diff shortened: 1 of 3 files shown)


tests/test_middleware_interface.py
```diff
@@ -362,6 +361,52 @@ def test_run_pipeline_empty_quantum_graph(self):
         with self.assertRaisesRegex(RuntimeError, "No data to process"):
             self.interface.run_pipeline(self.next_visit, {2})
 
+    def _assert_in_collection(self, butler, collection, dataset_type, data_id):
+        for dataset in butler.registry.queryDatasets(dataset_type, collections=collection, dataId=data_id):
+            return
+        self.fail(f"No datasets found matching {dataset_type}@{data_id} in {collection}.")
```
parejkoj, line 367: Would this work instead of the for loop: `self.assertIn(dataset_type, list(butler.registry.queryDatasets(...)))`? `for thing: return` tripped me up.

tests/test_middleware_interface.py
```diff
@@ -362,6 +361,52 @@ def test_run_pipeline_empty_quantum_graph(self):
...
+
+    def _assert_not_in_collection(self, butler, collection, dataset_type, data_id):
+        for dataset in butler.registry.queryDatasets(dataset_type, collections=collection, dataId=data_id):
+            self.fail(f"{dataset} matches {dataset_type}@{data_id} in {collection}.")
+
+    def test_clean_unsafe_datasets(self):
+        """Test that _clean_unsafe_datasets removes only "conflicting" datasets.
```
parejkoj, line 374: Worth noting here that "conflicting" datasets are anything without `detector` in their datasetType? I realize that's an implementation detail, but I had to look at it to understand what was "assert_in" and what was "assert_not_in" at the end.

parejkoj (approved):
Two small suggestions

tests/test_middleware_interface.py
kfindeisen: That requires reading in all the dataset refs, which, since this is the DECam dataset, can be sizeable.

tests/test_middleware_interface.py
kfindeisen: Well, the concept of a "conflicting" dataset is non-algorithmic: it's anything that would be produced with the same data ID (with or without different contents) for a different AP pipeline run. These just happen to be specific datasets I could think of that did/didn't meet those criteria.

Is there some other way to make the "in" vs. "not in" distinction more readable?

tests/test_middleware_interface.py
parejkoj: Yeah, I think a comment like that might help.

tests/test_middleware_interface.py
parejkoj: Maybe a note to that effect in this docstring. Something like this? `The "conflicting" datasets in this test are specific examples that clearly meet the "can be produced with the same data ID in different AP runs" criteria; currently "contains detector in the data ID".`

kfindeisen: I went with:

> A conflicting dataset is one that can be produced with the same data ID in different AP runs. This test currently uses `calibrate_config` and `src_schema` as examples of conflicting datasets, and `src` and `calexp` as counterexamples (which should not be removed).

PR #42, opened 2022-12-07
DM-37258: Logs getting dropped in prompt processing

kfindeisen:
This PR fixes and refactors the logging configuration to better emulate the behavior expected from `pipetask` and BPS. It also adds a log level environment variable, which was requested for debugging our slow run times.

### DIFF
--- a/python/activator/logger.py
+++ b/python/activator/logger.py
@@ -23,6 +23,9 @@ __all__ = ["GCloudStructuredLogFormatter", "setup_google_logger", "setup_usdf_lo
 
 import json
 import logging
+import os
+
+import lsst.daf.butler as daf_butler
 
 try:
     import lsst.log as lsst_log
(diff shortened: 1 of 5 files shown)


python/activator/logger.py
```diff
@@ -23,6 +23,61 @@
...
+    daf_butler.cli.cliLog.CliLog.setLogLevels(default_levels + _parse_log_levels(log_request))
+
+
+def _channel_all_to_pylog():
+    """Set up redirection of lsst.log and warning output to the Python logger.
+
+    This ensures that all service output is formatted consistently, making it easier to parse.
+    """
+    if lsst_log is not None:
+        lsst_log.configure_pylog_MDC("DEBUG", MDC_class=None)
```
hsinfang, line 77: Is what happens that the root logger gets set to DEBUG level here, and later the  `(None, "WARNING")` in `_set_lsst_logging_levels()` (line 63) resets it to WARNING?  So the level here doesn't really matter?

python/activator/logger.py
```diff
@@ -23,6 +23,61 @@
...
+
+def _set_lsst_logging_levels():
+    """Set up standard logging levels for LSST code.
+
+    This function consistently sets both the Python logger and lsst.log.
+    """
+    # Don't call CliLog.initLog because we need different formatters from what
+    # Butler assumes.
+    default_levels = [(None, "WARNING"),  # De-prioritize output from 3rd-party packages.
+                      ("lsst", "INFO"),   # Capture output from Middleware and pipeline.
```
hsinfang, line 64: Would it make sense to set the default `lsst.activator` log level here?

hsinfang (approved):

python/activator/logger.py
```diff
@@ -23,6 +23,61 @@
...
+    levels = [tuple(s.split("=", maxsplit=1)) for s in spec.split(None)]
+    return [(k if k != "." else None, v) for k, v in levels]
+
+
+def _set_lsst_logging_levels():
+    """Set up standard logging levels for LSST code.
+
+    This function consistently sets both the Python logger and lsst.log.
+    """
+    # Don't call CliLog.initLog because we need different formatters from what
```
timj, line 61: If there is demand I'm open to moving that code out of butler and into `utils` and allowing more configuration of things like the long log template.

python/activator/logger.py
kfindeisen: I'll admit I cargo-culted this from `daf_butler`, but it's possible to change the root logging level from the envvar, and I think we'd want the log4cxx-Python bridge to work regardless.

timj: You need to call this to get forwarding from C++ logs to python logging. I'm pretty sure that if you set this to INFO or WARNING then you will not be able to ever get DEBUG logs out of C++ (I might be wrong on this). The explicit setting of the log level later means that using DEBUG here is not a problem.

python/activator/logger.py
kfindeisen: I don't think there's demand, at least not yet. Maybe once we've got a clearer picture from commissioning about what we actually need in the service environment.

PR #43, opened 2022-12-15
DM-37382: fix repeating raw image 

hsinfang:


python/tester/upload_hsc_rc2.py
```diff
@@ -102,14 +102,17 @@ def get_hsc_visit_list(butler, n_sample):
     visits : `list` [`int`]
         A list of randomly selected visit IDs from the HSC-RC2 dataset
     """
+    if n_sample > 432:
+        raise ValueError(f"Requested {n_sample} groups, but only 432 are available.")
+
```
kfindeisen, line 107: I'm not sure this number needs to be hardcoded; what if the collection gets bigger (more data!) or smaller (rejected or failed processing)? I'd suggest either catching the `ValueError` raised by `random.sample`, or simply passing it through.

python/tester/upload_hsc_rc2.py
```diff
@@ -147,16 +147,16 @@ def prepare_one_visit(producer, group_id, butler, visit_id):
     )
 
     visits = set()
-    for dataId in refs.dataIds.expanded():
+    for data in refs.dataIds.expanded():
```
kfindeisen, line 150: `data` seems a bit confusing; it makes me think of a big blob of binary stuff. Why not `data_id`?

kfindeisen (approved):
Looks good, just minor comments.

python/tester/upload_hsc_rc2.py
hsinfang: This number is going to stay the same unless we have to a new dataset (e.g. RC3)  or when we change the implementation (e.g. allowing reuse of the same visit for different group). Because we are drawing the raw data, failed DRP processing doesn't matter either; DRP processing provides templates only. Nonetheless,  I agree it's better not to hardcode it. I feel it's nicer if the error message tells the user how many groups are available so the user can adjust for the next test.

PR #46, opened 2023-01-11
DM-37547: HSC exposure ID generator produces invalid IDs in 2023

kfindeisen:
This PR adds a regression test for the bug where HSC exposure IDs were being generated with values of 30,000,000+, and rewrites the ID generation algorithm to confine them in the allowed range.

### DIFF
--- a/python/tester/utils.py
+++ b/python/tester/utils.py
@@ -30,6 +30,15 @@ from astropy.io import fits
 _log = logging.getLogger("lsst." + __name__)
 _log.setLevel(logging.DEBUG)
 
+max_exposure = {
+    "HSC": 21474800,
+}
+"""A mapping of instrument to exposure_max (`dict` [`str`, `int`]).
+
+The values are copied here so we can access them without a Butler. All
+exposure IDs are in Middleware (integer) format, not native format.
+"""
+
 
 def get_last_group(bucket, instrument, date):
     """Identify the largest group number or a new group number.
--- a/python/tester/utils.py
+++ b/python/tester/utils.py
@@ -133,15 +133,18 @@ def make_hsc_id(group_num, snap):
 
     Notes
     -----
-    The current implementation allows up to 1000 group numbers per day.
-    It can overflow with ~60 calls to upload.py on the same day or
-    upload_hsc_rc2.py with a large N_GROUPS.
+    The current implementation gives illegal exposure IDs after September 2024.
+    If this generator is still needed after that, it will need to be tweaked.
     """
     # This is a bit too dependent on how group_num is generated, but I want the
     # group number to be discernible even after compressing to 8 digits.
-    night_id = (group_num // 100_000) % 2020_00_00     # Always 5 digits
+    year = (group_num // 100_00_00000) - 2023          # Always 1 digit, 0-1
+    night_id = (group_num % 100_00_00000) // 100000    # Always 4 digits up to 1231
     run_id = group_num % 100_000                       # Up to 5 digits, but usually 2-3
-    exposure_id = (night_id * 1000) + (run_id % 1000)  # Always 8 digits
+    exposure_id = (year*1200 + night_id) * 10000 + (run_id % 10000)  # Always 8 digits
+    if exposure_id > max_exposure["HSC"]:
+        raise RuntimeError(f"{group_num} translated to expId {exposure_id}, "
+                           f"max allowed is { max_exposure['HSC']}.")
     return f"HSCE{exposure_id:08d}", exposure_id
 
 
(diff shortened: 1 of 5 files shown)


tests/test_tester_utils.py
```diff
@@ -76,3 +81,22 @@ def test_get_last_group(self):
         # Test the case of no match
         last_group = get_last_group(bucket, "TestCam", "20110101")
         self.assertEqual(last_group, int(20110101) * 100_000)
+
+    def test_exposure_id_hsc(self):
+        group = "2023011100026"
```
ktlim, line 86: Why not just a number, avoiding the later `int()`?

tests/test_tester_utils.py
```diff
@@ -76,3 +81,22 @@ def test_get_last_group(self):
...
+        # Need a Butler registry to test ExposureIdInfo
+        with tempfile.TemporaryDirectory() as repo:
+            butler = butler_tests.makeTestRepo(repo)
+            HyperSuprimeCam().register(butler.registry)
+            for record in butler.registry.queryDimensionRecords("instrument", dataId={"instrument": "HSC"}):
```
ktlim, line 91: There should be only one, right?  Maybe check for that explicitly rather than iterating and taking only the last one.

tests/test_tester_utils.py
```diff
@@ -76,3 +81,22 @@ def test_get_last_group(self):
...
+                exp_max = record.exposure_max
+
+            _, str_exp_id, exp_id = make_exposure_id("HSC", int(group), 0)
+            butler_tests.addDataIdValue(butler, "visit", exp_id)
+            data_id = butler.registry.expandDataId({"instrument": "HSC", "visit": exp_id, "detector": 111})
+
+        self.assertEqual(str_exp_id, "HSCE%08d" % exp_id)
+        self.assertEqual(len(str_exp_id[4:]), 8)
```
ktlim, line 99: This is kind of redundant with the above.

python/tester/utils.py
```diff
@@ -121,18 +130,11 @@ def make_hsc_id(group_num, snap):
...
     """
-    # This is a bit too dependent on how group_num is generated, but I want the
-    # group number to be discernible even after compressing to 8 digits.
-    night_id = (group_num // 100_000) % 2020_00_00     # Always 5 digits
-    run_id = group_num % 100_000                       # Up to 5 digits, but usually 2-3
-    exposure_id = (night_id * 1000) + (run_id % 1000)  # Always 8 digits
+    # Use hash to distribute evenly over space of allowed IDs -- raw
+    # group ID has many digits that are almost always 0.
+    # hash returns an int64 on most systems, but HSC max is much smaller.
+    exposure_id = hash(group_num) % max_exposure["HSC"]
```
ktlim, line 137: The three problems with this are that it loses the monotonicity (which is not required but nice to have), it loses the correlation with the time of the test (also not required but nice to have), and it has potential collisions which cannot be predicted.

How about `((YYYY-2020)*1200+MMDD)*1000 + group_id % 1000`, which is good until the end of Oct 2037, by which time we had better be using real LSSTCam for testing?

tests/test_tester_utils.py
kfindeisen: We get a _lot_ of bugs related to integers being assigned as groups and then passed around like that, so I've insisted on other reviews that we always be strict with the types there. I could hardly make an exception for myself, especially since part of the problem is that we tend to _think_ of group IDs as integers when formally they're not.

tests/test_tester_utils.py
kfindeisen: Actually, it's not: if `exp_id` is 9+ digits, then the formatting (here or in the application code) will dutifully create a string with 9+ digits, even though it wouldn't get parsed by `astro_metadata_translator`. I can add a comment to that effect.

python/tester/utils.py
kfindeisen: Where does the factor of 1200 come from? I guess the earlier algorithm was essentially `(YYYY-2020)*10_000+MMDD`...

Though I do like it -- the date term is limited to 4 digits through 2027, and if we reserve 4 digits for the group ID then we're much less likely to overflow.

ktlim: 1200 is from 12 months per year and a two digit shift to clear the days.  If you want to have 9999 groups (4 digits), though, then you'd have to change the epoch year because you're limited to 1.75 years before you hit 21474800.

kfindeisen:
@ktlim, I implemented the revised generator (thanks for catching the 1.75-year caveat!). I'm still testing, but so far output looks reasonable:
```
DEBUG:lsst.activator.activator:Received Snap(instrument='HSC', detector=0, group='2023011200017', snap=0, exp_id=1120017, filter='HSC-G')
```

Can you take another look and let me know if the code meets with your approval?

ktlim (approved):
Everything looks awesome — except the commit comment, which says that monotonicity is sacrificed when it isn't now!

PR #47, opened 2023-01-19
DM-37072: Expand HSC calibs and templates in central repo

hsinfang:
This is a continuation of https://github.com/lsst-dm/prompt_prototype/pull/44 , to fix two bugs there.

### DIFF
--- a/bin.src/make_remote_butler.py
+++ b/bin.src/make_remote_butler.py
@@ -117,6 +117,11 @@ def _hsc_rc2(butler):
             "HSC/calib/DM-28636",
         ],
     )
+
+    butler.registry.registerCollection(
+        instrument.makeUnboundedCalibrationRunName(),
+        type=CollectionType.CHAINED
+    )
     butler.registry.setCollectionChain(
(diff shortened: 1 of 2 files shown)


bin.src/make_remote_butler.py
```diff
@@ -117,6 +117,11 @@ def _hsc_rc2(butler):
             "HSC/calib/DM-28636",
         ],
     )
+
+    butler.registry.registerCollection(
+        instrument.makeUnboundedCalibrationRunName(),
+        type=CollectionType.CHAINED
+    )
```
kfindeisen, line 124: Maybe add a similar line above for the calibration collection, just in case?

kfindeisen (approved):
Looks good, one minor suggestion.

bin.src/make_remote_butler.py
hsinfang: It wouldn't be needed because it's done already at https://github.com/lsst-dm/prompt_prototype/blob/ca1d53afe06bd239ac35211f4510781d0c23712d/bin.src/make_remote_butler.py#L95

PR #49, opened 2023-01-23
DM-36771: Bring Visit object in line with SAL spec

kfindeisen:
This PR significantly rewrites the `Visit` class to have matching fields to [the SAL `nextVisit` spec](https://ts-xml.lsst.io/sal_interfaces/ScriptQueue.html#nextvisit) (as modified for Kafka and Python). Since the `nextVisit` spec allows for states we hadn't previously considered in the prototype, I've also added checks or TODOs to the main codebase.


kfindeisen:
Tested with `d_2023_02_03` and the uploader workaround from #47.

python/activator/visit.py
```diff
@@ -1,24 +1,59 @@
...
+import enum
 
 
-@dataclass(frozen=True)
+@dataclass(frozen=True, kw_only=True)
 class Visit:
-    # elements must be hashable and JSON-persistable; built-in types recommended
-    instrument: str    # short name
+    # Elements must be hashable and JSON-persistable; built-in types
+    # recommended. list is not hashable, but gets special treatment because
```
bsmartradio, line 10: Should list be capitalized? Not entirely sure in this case.

bsmartradio (approved):
Everything looks good to me. I also double checked the files in case a changed variable was missed but didn't find anything.

python/activator/visit.py
```diff
@@ -1,24 +1,59 @@
...
+    # neither Kafka nor JSON deserialize sequences as tuples.
+
+    # Inherited from SAL next_visit schema; keep in sync with
+    # https://ts-xml.lsst.io/sal_interfaces/ScriptQueue.html#nextvisit
+    class CoordSys(enum.IntEnum):
+        # This is a redeclaration of lsst.ts.idl.enums.Script.MetadataCoordSys,
+        # but need Visit to work in code that can't import lsst.ts.
```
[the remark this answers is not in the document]
kfindeisen: I'll change it to "we need".

python/activator/visit.py
kfindeisen: I'd rather use the type's intrinsic capitalization; "List" could be confusing.

bsmartradio: That's what I thought, I just wasn't 100% sure.

PR #52, opened 2023-02-14
DM-37751: Add a script to make the export file for templates

hsinfang:


bin.src/make_template_export.py
```diff
@@ -0,0 +1,82 @@
...
+    parser.add_argument(
+        "--collection",
+        required=True,
+        help="The collection to query data to be exported.",
+    )
+    return parser
+
+
+def main():
+    logging.basicConfig(level=logging.DEBUG, stream=sys.stdout)
```
kfindeisen, line 53: Should this be `INFO` instead? I'm a bit worried that making the default logger `DEBUG` will lead to far too much information...

bin.src/make_template_export.py
```diff
@@ -0,0 +1,82 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+
+"""Export goodSeeingCoadd from a collection and make an export file
+for importing those data to a central prompt processing repository.
+"""
```
kfindeisen, line 26: Can you document that this takes _all_ coadds from the target collection, without selecting them in any way?

kfindeisen (approved):
Looks good! Only minor comments.

PR #53, opened 2023-03-02
DM-37745: Prompt processing preload fails silently if central repo doesn't have prerequisites

kfindeisen:
This PR replaces the old `_query_missing_datasets` with a new function, `_filter_datasets`, that raises if the source repository does not have anything matching the query. This behavior is needed in all contexts where `MiddlewareInterface` queries both the central and local repositories, though it's inappropriate for a general-purpose query.


isullivan (approved):
Looks very good. I appreciate the unit test testing the new functionality.

PR #56, opened 2023-03-28
DM-36162: Upgrade from SimplePipelineExecutor in prompt prototype

kfindeisen:
This PR redesigns `MiddlewareInterface` to be a per-exposure rather than a per-worker object. This change will make it much easier and more foolproof to work with `MiddlewareInterface` in the future, because it no longer needs to avoid storing per-exposure state. This done, the PR replaces `MiddlewareInterface`'s use of `SimplePipelineExecutor` with the `SeparablePipelineExecutor` introduced in lsst/ctrl_mpexec#230.

~The changes to `build-service.yml` and `Dockerfile.activator` are *not* part of the PR, and were temporary hacks to get a service container that had access to `SeparablePipelineExecutor`.~ I intend to wait until lsst/ctrl_mpexec#230 is incorporated into a nightly/weekly before merging this PR; this will allow all actions to complete without hacks.

### DIFF
--- a/python/activator/activator.py
+++ b/python/activator/activator.py
@@ -87,12 +87,6 @@ storage_client = boto3.client('s3', endpoint_url=s3_endpoint)
 central_butler = get_central_butler(calib_repo, instrument_name)
 # local_repo is a temporary directory with the same lifetime as this process.
 local_repo = make_local_repo(local_repos, central_butler, instrument_name)
-# Initialize middleware interface.
-mwi = MiddlewareInterface(central_butler,
-                          image_bucket,
-                          instrument_name,
-                          skymap,
-                          local_repo.name)
 
 
 def check_for_snap(
@@ -235,6 +229,13 @@ def next_visit_handler() -> Tuple[str, int]:
             f"Expected {instrument_name}, received {expected_visit.instrument}."
         expid_set = set()
 
+        # Create a fresh MiddlewareInterface object to avoid accidental
+        # "cross-talk" between different visits.
+        mwi = MiddlewareInterface(central_butler,
+                                  image_bucket,
+                                  instrument_name,
+                                  skymap,
+                                  local_repo.name)
         # Copy calibrations for this detector/visit
         mwi.prep_butler(expected_visit)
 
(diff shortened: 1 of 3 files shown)


python/activator/activator.py
```diff
@@ -84,12 +84,15 @@
 
 storage_client = boto3.client('s3', endpoint_url=s3_endpoint)
 
+central_butler = get_central_butler(calib_repo, instrument_name)
+# local_repo is a temporary directory with the same lifetime as this process.
+local_repo = make_local_repo(local_repos, central_butler, instrument_name)
```
TallJimbo, line 89: I can image why we can't use a context manager for cleanup here; do we need to worry about that, or is the lifetime of this process tied to some kind of more complete reset of the compute resource that will take care of that?

python/activator/middleware_interface.py
```diff
@@ -358,9 +383,10 @@ def prep_butler(self, visit: Visit) -> None:
         wcs = self._predict_wcs(detector, visit)
         center, radius = self._detector_bounding_circle(detector, wcs)
 
-        # central repo may have been modified by other MWI instances.
+        # repos may have been modified by other MWI instances.
         # TODO: get a proper synchronization API for Butler
         self.central_butler.registry.refresh()
+        self.butler.registry.refresh()
```
TallJimbo, line 389: I'm curious what this `refresh` and the one below are doing for you, especially after the switch to per-visit `MiddlewareInterface` objects.  Do workers ever need to use something that is only added to the repo by some other worker?  Or is this to avoid races when creating some shared repo content?

TallJimbo (approved):

python/activator/activator.py
kfindeisen: For performance reasons, we want to reuse the same local repo for processing multiple exposures. And AFAIK Flask doesn't let us pass in arguments to `next_visit_handler`, so it needs to be (directly or indirectly) a global. I think that precludes use of a context manager.

Basically, the way the Prompt Processing architecture is set up, the Python process represented by `activator.py` has the same lifetime as the (virtual) filesystem it's running on. I don't know if that answers your second question.

python/activator/middleware_interface.py
kfindeisen: Oh, sorry, I misread the original question. Yes, the `central_butler` refreshes are to avoid conflicts between concurrently writing workers. So far the only such conflicts have involved things like collection creation, and there shouldn't be any right now, but as with the local repo I'm trying to be robust to unknown future changes.

PR #57, opened 2023-04-12
DM-38405: Update the nextVisit events produced by the tester upload scripts

hsinfang:


python/tester/upload.py
```diff
@@ -46,31 +42,32 @@ class Instrument:
...
     """Simulate the observation of a single on-sky pointing.
 
     Parameters
     ----------
-    producer : `confluent_kafka.Producer`
-        The client that posts ``next_visit`` messages.
-    visit_infos : `set` [`activator.Visit`]
+    kafka_url : `str`
+        The URL of the Kafka REST Proxy to send ``next_visit`` messages to.
+    visit_infos : `set` [`activator.FannedOutVisit`]
```
kfindeisen, line 52: Why use `FannedOutVisit` here? I would have thought the uploaders could just use `BareVisit` internally.

python/activator/visit.py
```diff
@@ -48,7 +48,16 @@ class Dome(enum.IntEnum):
     survey: str                 # survey name
```
kfindeisen, line 48: Can you also update the documentation of `filters` to say that these filter names are in the Middleware format, including `~` for composite filters? Following up on https://lsstc.slack.com/archives/C0345PUU4N9/p1679429952829539.

kfindeisen (commented):
One minor comment and one request:

python/activator/visit.py
```diff
@@ -48,7 +48,16 @@ class Dome(enum.IntEnum):
...
+    def __str__(self):
+        """Return a short string that represents the visit but does not
+        include complete metadata.
+        """
+        return f"(groupId={self.groupId}, salIndex={self.salIndex})"
+
+
+@dataclass(frozen=True, kw_only=True)
+class FannedOutVisit(BareVisit):
+    # Extra information is added by the fan-out serviuce at USDF.
```
ebellm, line 60:
```suggestion
    # Extra information is added by the fan-out service at USDF.
```

ebellm (approved):

python/tester/upload.py
hsinfang: Not really because how `raw_pool` is unless we rewrite it completely.  The thing is that we still want to keep track of files at the detector level and hence `FannedOutVisit` is quite convenient for it.  If we use `BareVisit` we'd still need those detector IDs somehow. It seems doable but I'm not sure if that'd really be better.

PR #61, opened 2023-05-04
DM-38596: Get bucket_topic from env var.

ktlim:


Dockerfile.activator
```diff
@@ -9,6 +9,7 @@ ARG PORT
 ARG CALIB_REPO
 ARG IMAGE_BUCKET
 ARG IMAGE_TIMEOUT
+ARG BUCKET_TOPIC
```
hsinfang, line 12: Not about this PR, but this makes me wonder why some env vars in activator aren't here and it works just fine...

hsinfang (approved):

Dockerfile.activator
hsinfang: We probably should remove them then. I can create a ticket for that.

hsinfang: [DM-39030](https://jira.lsstcorp.org/browse/DM-39030)

PR #64, opened 2023-05-16
DM-39188: Export raws only if they are not already in the central repo

hsinfang:
I hesitate about this because adding a query with the central repo means time spent waiting for the database query, and it probably won't scale well.  If we can agree just don't export raws in all cases, it seems better just to remove it.  Do we really want to wait for [DM-36051](https://jira.lsstcorp.org/browse/DM-36051).?

### DIFF
--- a/python/activator/middleware_interface.py
+++ b/python/activator/middleware_interface.py
@@ -759,7 +759,7 @@ class MiddlewareInterface:
                   f"detector {self.visit.detector} of {exposure_ids}.")
 
     def export_outputs(self, exposure_ids: set[int]) -> None:
-        """Copy raws and pipeline outputs from processing a set of images back
+        """Copy pipeline outputs from processing a set of images back
         to the central Butler.
 
         Parameters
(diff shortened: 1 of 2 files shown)


ktlim (approved):

hsinfang:
Change of mind after some Slack discussions.  We can just not export raws at all.

hsinfang:
Thanks for catching that. I'm adding it to `test_export_outputs` but indeed they aren't really run in parallel and I'm not sure how to enforce that in the current framework.

hsinfang:
As a `multiprocessing` newbie.... if I just use it from inside a test to do a dummy test, it can pass. But using the middleware interface gave me `TypeError: cannot pickle 'lsst.afw.cameraGeom._cameraGeom.Camera' object` . 
Will see whether I can avoid sending non-pickleble objects...

kfindeisen:
Sorry, I forgot that `multiprocessing` needs everything to be picklable. 😞 Since the actual framework never tries to share `MiddlewareInterface` objects, it's probably not worth designing them for it...

PR #68, opened 2023-05-25
DM-39395: Use day_obs in run names

kfindeisen:
This PR changes the naming convention for output runs so that all products of the same pipeline and the same night should go into the same run.


hsinfang (approved):
Looks great!

PR #72, opened 2023-06-15
DM-39678: Allow Prompt Prototype to be configured with no pipeline

kfindeisen:
This PR adds some logic that lets us skip Prompt Processing execution for specific types of visits. This is mainly intended for spectroscopic observations taken with LATISS.


dspeck1 (approved):
Looks good.  Only comment is duplicate of other review in slaclab repo around if it  appropriate to move to structured config file instead of space delimited file and regex parsing?  If that is normal convention that is fine.

PR #75, opened 2023-07-21
DM-39655: Use a list of "fallback" pipelines in Prompt Processing

kfindeisen:
This PR adds support for multiple pipelines that may be attempted for the same visit, in sequence. This required a significant reorganization of `MiddlewareInterface`, moving decisions that depend on the choice of pipeline to as late in the workflow as possible. However, the new code has fewer implicit dependencies, making it more robust to future changes.

### DIFF
--- a/python/activator/middleware_interface.py
+++ b/python/activator/middleware_interface.py
@@ -616,24 +616,25 @@ class MiddlewareInterface:
         """Pre-register output collections in advance of running the pipeline.
         """
         self.butler.registry.refresh()
-        pipeline_file = self._get_pipeline_file()
-        self.butler.registry.registerCollection(
-            self._get_init_output_run(pipeline_file, self._day_obs),
-            CollectionType.RUN)
-        self.butler.registry.registerCollection(
-            self._get_output_run(pipeline_file, self._day_obs),
-            CollectionType.RUN)
+        for pipeline_file in self._get_pipeline_files():
+            self.butler.registry.registerCollection(
+                self._get_init_output_run(pipeline_file, self._day_obs),
+                CollectionType.RUN)
+            self.butler.registry.registerCollection(
+                self._get_output_run(pipeline_file, self._day_obs),
+                CollectionType.RUN)
 
-    def _get_pipeline_file(self) -> str:
-        """Identify the pipeline to be run, based on the configured instrument
+    def _get_pipeline_files(self) -> str:
+        """Identify the pipelines to be run, based on the configured instrument
         and visit.
 
         Returns
         -------
-        pipeline : `str`
-            A path to a configured pipeline file.
+        pipelines : sequence [`str`]
+            A sequence of paths to a configured pipeline file, in order from
+            most preferred to least preferred.
         """
-        return self.pipelines.get_pipeline_file(self.visit)
+        return [self.pipelines.get_pipeline_file(self.visit)]
 
     def _prep_pipeline(self, pipeline_file) -> lsst.pipe.base.Pipeline:
         """Setup the pipeline to be run, based on the configured instrument and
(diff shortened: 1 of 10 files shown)


python/activator/middleware_interface.py
```diff
@@ -740,20 +741,42 @@ def run_pipeline(self, exposure_ids: set[int]) -> None:
...
+                skip_existing_in=None
+            )
+            qgraph = executor.make_quantum_graph(pipeline, where=where)
+            if len(qgraph) == 0:
+                # Diagnostic logs are the responsibility of GraphBuilder.
+                _log.error(f"Could not build quantum graph for {pipeline_file}; "
+                           "see previous logs for details.")
+                continue
+            # Past this point, partial execution creates datasets.
+            # Don't retry -- either fail (raise) or break.
```
hsinfang, line 766: Do I understand correctly that for now, the only scenario to fallback is a failure in making the quantum graph?

hsinfang (approved):

python/activator/middleware_interface.py
```diff
@@ -776,16 +787,23 @@ def export_outputs(self, exposure_ids: set[int]) -> None:
...
+            output_run = self._get_output_run(pipeline_file, self._day_obs)
+            exports = self._export_subset(exposure_ids,
+                                          # TODO: find a way to merge datasets like *_config
+                                          # or *_schema that are duplicated across multiple
+                                          # workers.
+                                          self._get_safe_dataset_types(self.butler),
+                                          in_collections=output_run,
+                                          )
+            if exports:
+                exported = True
```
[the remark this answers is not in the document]
kfindeisen: I don't expect more than one collection to contain outputs, but I'm not confident that that will always be true in the presence of failures or future code changes (e.g., doing pre-execution as a separate step, or expanding fallback to more situations). This way seems safer.

python/activator/middleware_interface.py
kfindeisen: Yes, at least for now. Do you know of any other scenarios where this is a good strategy?

hsinfang: Got another one that might be relevant here: quantum graph builds fine, but the needed datasets aren't really available so it fails. It can show as something like
`FileNotFoundError: Not enough datasets (0) found for non-optional connection`...

kfindeisen: I assume this happens only with `PrerequisiteInput`? I.e., it does not affect templates?

hsinfang: Looks like only with `PrerequisiteInput`  
https://github.com/lsst/pipe_base/blob/main/python/lsst/pipe/base/connections.py#L852-L855

I haven't traced down the middleware enough to understand whether the qgraph would be built successfully in this case (I don't know when `adjustQuantum` happens).  When I encountered this error, I don't think the PP fallback pipeline ran, so I'd guess the qgraph was built.

kfindeisen: It looks like, despite the docstring, `adjustQuantum` is actually run [right before the quantum is executed](https://github.com/lsst/ctrl_mpexec/blob/8caecd393392dbc7d64e955c69b1c8bb9059ce55/python/lsst/ctrl/mpexec/singleQuantumExecutor.py#L431-L433).

hsinfang: Ah, the qgraph building would fail. But it's not generating an empty qgraph which is the case the PP fallback pipeline is designed for.

kfindeisen: Sorry for the confusion on Slack! I've filed [DM-41596](https://jira.lsstcorp.org/browse/DM-41596) to address the bug.

PR #77, opened 2023-07-31
DM-39514: Add validity-range workaround to prep_butler

kfindeisen:
This PR uses `queryDatasetAssociations` to filter out expired calibs during local repo setup. This is not a foolproof solution; in particular, error detection is quite poor because the current code needs to be agnostic to exactly which calibs the pipeline(s) need(s). Error detection will be improved once [DM-40245](https://jira.lsstcorp.org/browse/DM-40245) rewrites calib loading to be more dynamic.

### DIFF
--- a/python/activator/middleware_interface.py
+++ b/python/activator/middleware_interface.py
@@ -505,6 +505,8 @@ class MiddlewareInterface:
         # TODO: we can't filter by validity range because it's not
         # supported in queryDatasets yet.
         calib_where = f"detector={detector_id} and physical_filter='{filter}'"
+        # private_sndStamp is in TAI, not UTC, but difference shouldn't matter
+        calib_date = datetime.datetime.fromtimestamp(self.visit.private_sndStamp, tz=datetime.timezone.utc)
         # TODO: we can't use findFirst=True yet because findFirst query
         # in CALIBRATION-type collection is not supported currently.
         calibs = set(_filter_datasets(
@@ -512,7 +514,9 @@ class MiddlewareInterface:
             ...,
             collections=self.instrument.makeCalibrationCollectionName(),
             instrument=self.instrument.getName(),
-            where=calib_where))
+            where=calib_where,
+            calib_date=calib_date,
+        ))
         if calibs:
             for dataset_type, n_datasets in self._count_by_type(calibs):
                 _log.debug("Found %d new calib datasets of type '%s'.", n_datasets, dataset_type)
@@ -918,8 +922,11 @@ class _MissingDatasetError(RuntimeError):
     pass
 
 
-def _filter_datasets(src_repo: Butler, dest_repo: Butler,
-                     *args, **kwargs) -> collections.abc.Iterable[lsst.daf.butler.DatasetRef]:
+def _filter_datasets(src_repo: Butler,
+                     dest_repo: Butler,
+                     *args,
+                     calib_date: datetime.datetime | None = None,
+                     **kwargs) -> collections.abc.Iterable[lsst.daf.butler.DatasetRef]:
     """Identify datasets in a source repository, filtering out those already
     present in a destination.
 
@@ -932,6 +939,10 @@ def _filter_datasets(src_repo: Butler, dest_repo: Butler,
         The repository in which a dataset must be present.
     dest_repo : `lsst.daf.butler.Butler`
         The repository in which a dataset must not be present.
+    calib_date : `datetime.datetime`, optional
+        If provided, also filter anything other than calibs valid at
+        ``calib_date`` and check that at least one valid calib was found.
+        Any ``datetime`` object must be aware.
     *args, **kwargs
         Parameters for describing the dataset query. They have the same
         meanings as the parameters of `lsst.daf.butler.Registry.queryDatasets`.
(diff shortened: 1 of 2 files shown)


python/activator/middleware_interface.py
```diff
@@ -1010,3 +1012,43 @@ def _remove_from_chain(butler: Butler, chain: str, old_collections: collections.
...
+    """Trim a set of calib datasets to those that are valid at a particular time.
+
+    Parameters
+    ----------
+    butler : `lsst.daf.butler.Butler`
+        The Butler to query for validity data.
+    collections : collection expression
+        The calibration collection(s), or chain(s) containing calibration
+        collections, to query for validity data.
+    unfiltered_calibs : collection [`lsst.daf.butler.DatasetRef`]
```
TallJimbo, line 1053: In middleware packages we've started writes types like this in docs as
```
~collections.abc.Collection` [`lsst.daf.butler.DatasetRef`]
```
to make the container type show up as a link to the standard library docs.  You don't need to adopt that here - I don't think we've made any attempt to formalize this practice - but it seemed worth mentioning in case you find that you prefer that approach, too.

python/activator/middleware_interface.py
```diff
@@ -1010,3 +1012,43 @@ def _remove_from_chain(butler: Butler, chain: str, old_collections: collections.
...
+        associations.update(
+            (a.ref, a) for a in butler.registry.queryDatasetAssociations(
+                dataset_type, collections, collectionTypes={CollectionType.CALIBRATION}, flattenChains=True
+            )
+        )
+
+    t = astropy.time.Time(date, scale='utc')
+    # DatasetAssociation.timespan guaranteed not None
+    return [ref for ref in unfiltered_calibs
+            if ref in associations and associations[ref].timespan.contains(t)]
```
TallJimbo: If more than one CALIBRATION-type collection is passed in (either directly as part of a CHAINED collection), this can return more than one `DatasetRef` of the same type and data ID (there's no logic like `findFirst=True` argument to `queryDatasets` happening here).  It looks like that's okay, since you export all of the collections you'll want to search, too, and that will export the search order so later lookups do the right thing, but I wanted to check.

TallJimbo (approved):

python/activator/middleware_interface.py
```diff
@@ -932,6 +939,10 @@ def _filter_datasets(src_repo: Butler, dest_repo: Butler,
         The repository in which a dataset must be present.
     dest_repo : `lsst.daf.butler.Butler`
         The repository in which a dataset must not be present.
+    calib_date : `datetime.datetime`, optional
+        If provided, also filter anything other than calibs valid at
+        ``calib_date`` and check that at least one valid calib was found.
+        Any ``datetime`` object must be aware.
```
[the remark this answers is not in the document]
kfindeisen: I don't see anything missing?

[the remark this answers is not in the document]
TallJimbo: What does it mean for a datetime object to be "aware"?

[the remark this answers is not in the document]
kfindeisen: https://docs.python.org/3/library/datetime.html#aware-and-naive-objects

python/activator/middleware_interface.py
kfindeisen: Yes, this has actually already happened (test runs picked up multiple `sky` datasets with overlapping validity ranges). Though they also had different runs...

Since ([according to a TODO](https://github.com/lsst-dm/prompt_prototype/blob/tickets/DM-39514/python/activator/middleware_interface.py#L513-L514)) we can't use `findFirst` with calibs, is there some other way to get `_export_calibs` to return unique calibs?

TallJimbo: `queryDatasetAssociations` will return results in the right order (I don't think this is documented but I'd be happy to make it a guarantee), so after you restrict the results to one instant in time, you could deduplicate manually on dataset type + data ID by using the first one returned.

The only other option at present is to use `Registry.findDataset` on one dataset type + data ID combination at a time.

kfindeisen: I don't think deduplication should be in scope for `_filter_calibs_by_date`, since the only way it can produce duplicate output is if it has duplicate input; it's not an artifact of using `queryDatasetAssociations` internally.

As for `findDataset`... I guess that would still be a net gain?

kfindeisen: Actually, on second thought, wouldn't using `findDataset` make `queryDatasetAssociations` unnecessary? 🤔

TallJimbo: It _might_ be a net gain. My understanding is that with this PR's changes all calibration lookups are still correct, but you export more than you have to (especially in the limit that there are many unnecessary CALIBRATION collections in the chain).  With `findDataset` you'd be doing many small database queries (to the point where the per-query latency overhead can become a performance problem; this is what makes step1 QG generation slow).  I don't know if that's a good trade either overall or on the basis of where the time is spent.

I also think I saw a comment saying that you don't know which dataset types will be needed up-front, and that probably rules out going with `findDataset` right now anyway.

kfindeisen: I thought the proposal was to use `findDataset` to filter the results of `queryDatasets`, in which case I have a (superset of) the dataset types already.

TallJimbo: Yes, I was thinking of `findDataset` as an alternative to `queryDatasetAssociations`.

TallJimbo: Yes, doing a full `queryDatasets` with `findFirst=False` to find the candidate dataset refs and then following up with `findDatasets` would indeed work.  If time spent here in `prep_butler` is "cheap" compared to time spent doing the export or import, that's a pretty good plan.

PR #78, opened 2023-08-03
DM-40210: Guard against future ap_pipe changes

kfindeisen:
This PR adds an exclusion of the `rbClassify` task from the Prompt Processing pipeline, in case it is added to `ap_pipe` in the future (see comments added in lsst/ap_verify#196 for details on the blockers). This will prevent us from accidentally trying to run the task before we are ready for it.


rai-harshit (approved):
Looks good.

PR #80, opened 2023-08-25
DM-40511: Refresh the unit test data repo

hsinfang:


tests/data/make_central_repo.sh
```diff
@@ -4,24 +4,24 @@
...
+butler collection-chain central_repo refcats refcats/DM-28636
+butler collection-chain central_repo DECam/templates u/elhoward/DM-38243/templates/20230320T214145Z
 butler collection-chain central_repo DECam/defaults DECam/calib,refcats,DECam/templates,skymaps
 
 # Empty out files and make them size 0 in the registry.
 # We do not empty the camera description in ``DECAM/calib/unbounded`
 # because we need it to load the camera geometry.
 find central_repo/refcats -name "*.fits" -execdir sh -c '> "$1"' -- {} \;
-find central_repo/templates -name "*.fits" -execdir sh -c '> "$1"' -- {} \;
+find central_repo/*/*/*/templates -name "*.fits" -execdir sh -c '> "$1"' -- {} \;
```
kfindeisen, line 19: I don't know if this will work with `find`, but if so it should be more robust:
```suggestion
find central_repo/**/templates -name "*.fits" -execdir sh -c '> "$1"' -- {} \;
```

pipelines/calibrate.py
```diff
@@ -1,13 +1,16 @@
+# TODO DM-40388: Remove this file, once the new pipeline is vetted.
```
kfindeisen, line 1: This comment refers to an ongoing project to replace `CalibrateTask` with `CalibrateImageTask` in the `ApVerify` pipeline. Do we have a "new pipeline" in PP?

pipelines/calibrate.py
kfindeisen, line 1: It looks like this file [is exclusive to the DECam test pipeline](https://github.com/search?q=repo%3Alsst-dm%2Fprompt_prototype%20calibrate.py&type=code), despite not being in an instrument-specific directory. Could you move it to `tests/data`?

pipelines/calibrate.py
```diff
@@ -1,13 +1,16 @@
...
 
 config.connections.astromRefCat = "gaia_dr2_20200414"
 config.astromRefObjLoader.anyFilterMapsToThis = "phot_g_mean"
-config.astromRefObjLoader.filterMap = {}
 
 # Use panstarrs for photometry (grizy filters).
 config.connections.photoRefCat = "ps1_pv3_3pi_20170110"
+config.photoRefObjLoader.filterMap = {
+    "u": "g",
+    "VR": "g"}
```
kfindeisen, line 16: Should this commit be merged with "Refresh the unit test data repo", or did the tests work fine without these changes?

tests/data/ApPipe.yaml
```diff
@@ -1,12 +1,11 @@
 description: End to end Alert Production pipeline specialized for HiTS-2015
 #
-# This file was copied from ap_verify_ci_hits2015 branch DM-34237 and modified
+# This file was copied from ap_verify_ci_hits2015/pipelines/ApPipe.yaml
+# with doSolarSystemAssociation disabled and modified
```
kfindeisen, line 4: There's a bit of ambiguous parsing here. Perhaps "and modified, with solar system association disabled"?

Out of curiosity, what other modifications were made?

kfindeisen (approved):
Looks good, just some self-consistency questions.

pipelines/calibrate.py
kfindeisen: Ugh, I guess we have to, since `ApPipe.yaml` imports the main one. Perhaps "Remove this file when `ApPipe` is migrated to `CalibrateImageTask`"?

hsinfang: From reading the tickets it looks to me that eventually `CalibrateImageTask` will replace `CalibrateTask` everywhere, so when that happens I assume we'd want PP to follow too?  DM-40388 looks like the right ticket for that, but it can be a separate ticket too. Thanks for adding a comment in DM-40388.  

Tagging @isullivan  @ebellm

pipelines/calibrate.py
hsinfang: Tests work without these changes.

tests/data/ApPipe.yaml
hsinfang: Bad english. This file is the same as `ap_verify_ci_hits2015/pipelines/ApPipe.yaml` except 
1) `doSolarSystemAssociation` is `false` instead of `true` 
2)  `$PROMPT_PROTOTYPE_DIR/pipelines/calibrate.py` instead of `$AP_VERIFY_CI_HITS2015_DIR/config/calibrate.py` . 
I'll rephrase.

pipelines/calibrate.py
ebellm: We certainly want the PP payload in sync with `ap_pipe` and `ap_verify`.

tests/data/make_central_repo.sh
hsinfang: That doesn't work, but this works 
```find central_repo -name "*templates*fits" -execdir sh -c '> "$1"' -- {} \;```
and will use this.

PR #81, opened 2023-08-29
DM-35828: Use MDC or equivalent in prompt_prototype logging

kfindeisen:
This PR adds a context manager that allows all logging statements (including those from other packages) to be decorated with metadata about the visit-detector being processed. It also adds a JSON-based log formatter that makes it easier for Grafana queries to take advantage of the extra information.

### DIFF
--- a/python/activator/logger.py
+++ b/python/activator/logger.py
@@ -19,7 +19,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-__all__ = ["GCloudStructuredLogFormatter", "setup_google_logger", "setup_usdf_logger",
+__all__ = ["GCloudStructuredLogFormatter", "UsdfJsonFormatter", "setup_google_logger", "setup_usdf_logger",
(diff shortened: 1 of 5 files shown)


python/activator/logger.py
```diff
@@ -171,17 +171,62 @@ def __init__(self, labels=None):
...
+            self._labels = {}
+
+    def format(self, record):
+        # format updates record.message, but the full info is *only* in the return value.
+        msg = super().format(record)
+
+        # Many LogRecord attributes are only useful for interrogating the
+        # record in Python; filter to what's useful in Grafana.
+        entry = {
+            # formatTime only automatically handles msecs with the default format???
```
hsinfang, line 212: Is it intentional to leave the `???` ?

hsinfang (approved):
I know quite little to say about the thread-safe design. But overall it looks good to me, and this will help a lot in searching logs in Loki.

python/activator/logger.py
kfindeisen: You're right, that was rude of me. I'll reword.

PR #84, opened 2023-09-14
DM-37387: A tester script using LATISS data in /repo/embargo

hsinfang:
A first part of this ticket to get test data into `s3://rubin:rubin-pp-users/central_repo`

### DIFF
--- a/bin.src/make_latiss_export.py
+++ b/bin.src/make_latiss_export.py
@@ -1,0 +1,107 @@
+#!/usr/bin/env python
+# This file is part of prompt_prototype.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+
+"""Selectively export the contents of the LATISS dataset.
+
+This script selects some LATISS data in a source butler repo, and makes an export
+file for importing to the test central prompt processing repository.
+"""
+
+
+import argparse
+import logging
+import sys
+
+import lsst.daf.butler as daf_butler
+from lsst.utils.timer import time_this
+
+
+def _make_parser():
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--src-repo",
+        required=True,
+        help="The location of the repository from which datasets are exported.",
+    )
+    return parser
+
+
+def main():
+    logging.basicConfig(level=logging.INFO, stream=sys.stdout)
+
+    args = _make_parser().parse_args()
+    src_butler = daf_butler.Butler(args.src_repo)
+
+    with time_this(msg="Datasets and collections exported", level=logging.INFO):
+        _export_for_copy(src_butler)
+
+
+def _export_for_copy(butler):
+    """Export selected data to make copies in another butler repo.
+
+    Parameters
+    ----------
+    butler : `lsst.daf.butler.Butler`
+        The source Butler from which datasets are exported.
+    """
+    with butler.export(format="yaml") as contents:
+        logging.debug("Selecting deepCoadd datasets")
+        records = butler.registry.queryDatasets(
+            datasetType="deepCoadd",
+            collections="LATISS/runs/AUXTEL_DRP_IMAGING_2023-07AB-05AB/"
+                        "w_2023_19/PREOPS-3598/20230726T202836Z",
+        )
+        contents.saveDatasets(records)
+
+        logging.debug("Selecting refcats datasets")
+        records = butler.registry.queryDatasets(datasetType=..., collections="refcats")
+        contents.saveDatasets(records)
+
+        logging.debug("Selecting skymaps dataset")
+        records = butler.registry.queryDatasets(
+            datasetType="skyMap", collections="skymaps", dataId={"skymap": "latiss_v1"}
+        )
+        contents.saveDatasets(records)
+
+        logging.debug("Selecting datasets in LATISS/calib")
+        records = butler.registry.queryDatasets(
+            datasetType=..., collections="LATISS/calib"
+        )
+        contents.saveDatasets(records)
+
+        # Save selected collections and chains
+        for collection in butler.registry.queryCollections(
+            expression="LATISS/calib",
+            flattenChains=True,
+        ) + [
+            "LATISS/calib",
+            "LATISS/calib/DM-36719",
+            "LATISS/calib/DM-38946",
+            "LATISS/calib/DM-39505",
+            "LATISS/templates",
+        ]:
+            contents.saveCollection(collection)
+
+
+if __name__ == "__main__":
+    main()
--- a/bin.src/make_latiss_export.py
+++ b/bin.src/make_latiss_export.py
@@ -31,9 +31,12 @@ file for importing to the test central prompt processing repository.
 import argparse
 import logging
 import sys
+import tempfile
 
 import lsst.daf.butler as daf_butler
 from lsst.utils.timer import time_this
+
+from activator.middleware_interface import _filter_datasets
 
 
 def _make_parser():
@@ -42,6 +45,14 @@ def _make_parser():
         "--src-repo",
         required=True,
         help="The location of the repository from which datasets are exported.",
+    )
+    parser.add_argument(
+        "--target-repo",
+        required=False,
+        help="The location of the repository to which datasets are exported. "
+             "Datasets already existing in the target repo will not be "
+             "exported from the source repo. If no target repo is given, all "
+             "selected datasets in the source repo will be exported.",
     )
     return parser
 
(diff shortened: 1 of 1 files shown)


hsinfang:
An obvious problem here is that some export scripts are somewhat similar but I haven't bothered to consolidate them. In particular this new script and `make_hsc_rc2_export.py` are quite close and I probably should combine them to one.

bin.src/make_latiss_export.py
```diff
@@ -103,7 +122,10 @@ def _export_for_copy(butler):
             "LATISS/calib/DM-39505",
             "LATISS/templates",
         ]:
-            contents.saveCollection(collection)
+            try:
+                target_butler.registry.queryCollections(collection)
+            except (AttributeError, daf_butler.registry.MissingCollectionError):
+                contents.saveCollection(collection)
```
kfindeisen, line 128: Doesn't [queryCollections](https://pipelines.lsst.io/v/daily/py-api/lsst.daf.butler.Registry.html#lsst.daf.butler.Registry.queryCollections) return an empty sequence if the collection is absent?

What happens if `target_butler` is `None`?

bin.src/make_latiss_export.py
```diff
@@ -0,0 +1,110 @@
...
+        for collection in butler.registry.queryCollections(
+            expression="LATISS/calib",
+            flattenChains=True,
+        ) + [
+            "LATISS/calib",
+            "LATISS/calib/DM-36719",
+            "LATISS/calib/DM-38946",
+            "LATISS/calib/DM-39505",
+            "LATISS/templates",
+        ]:
```
kfindeisen, line 124: Should `refcats` also be in this list? That should make it unnecessary to edit the chain by hand.

kfindeisen (changes requested):
The initial script looks very clean and clear. However, I don't think I understand the `--target-repo` argument or the code that depends on it. Could you please clarify how it works, and when the caller might opt not to use it?

bin.src/make_latiss_export.py
```diff
@@ -0,0 +1,110 @@
...
+    end = time.time_ns()
+    logging.info("Export finished in %.3fs.", 1e-9 * (end - start))
+
+
+def _export_for_copy(butler):
+    """Export selected data to make copies in another butler repo.
+
+    Parameters
+    ----------
+    butler: `lsst.daf.butler.Butler`
```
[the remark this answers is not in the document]
hsinfang: Thank you! I keep forgetting about the extra space.

bin.src/make_latiss_export.py
```diff
@@ -74,23 +85,31 @@ def _export_for_copy(butler):
             collections="LATISS/runs/AUXTEL_DRP_IMAGING_2023-07AB-05AB/"
             + "w_2023_19/PREOPS-3598/20230726T202836Z",
         )
-        contents.saveDatasets(records)
+        for record in records:
+            if not target_butler or not target_butler.exists(record):
+                contents.saveDatasets({record})
```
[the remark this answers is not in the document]
hsinfang: I was thinking to reserve the possibility of not having a `target repo`, in the case that one may want to create a fresh new repo to dump everything into. 

You are right, doing one query will definitely be more efficient. I'll look into the MiddlewareInterface code.

bin.src/make_latiss_export.py
hsinfang: An empty sequence is returned If nothing matches the given collection wildcard expression. But, a 
```MissingCollectionError: No collection with name 'specific_collection' found.```  
if that `specific_collection` doesn't exist in that butler.

Comment (hsinfang) [reply bin.src/make_latiss_export.py:128]: And `AttributeError` if `target_butler` is `None`

**Comment by hsinfang** [comment]
This script makes an `export.yaml` file and this export file may be used to populate an empty freshly created butler repo, or to add more datasets into an existing repo.  The motivation I added the `--target-repo` argument was for the second case, and I wanted to skip the datasets that already exist in the target repo (`s3://rubin:rubin-pp-users/central_repo` to be precise).  It turns out that the only existing datasets are `refcats` for what I'm doing in this ticket.  That's expected; the script as-it was used to verify that.  
I think I'm struggling a bit between making the script more generic versus for the very specific usage in mind. The script assumes a lot about the data already. So maybe checking every dataset doesn't make sense.

File: bin.src/make_latiss_export.py [hunk 5]
```diff
@@ -0,0 +1,110 @@
...
+    ----------
+    butler: `lsst.daf.butler.Butler`
+        The source Butler from which datasets are exported.
+    """
+    with butler.export(format="yaml") as contents:
+        logging.debug("Selecting deepCoadd datasets")
+        records = butler.registry.queryDatasets(
+            datasetType="deepCoadd",
+            collections="LATISS/runs/AUXTEL_DRP_IMAGING_2023-07AB-05AB/"
+            + "w_2023_19/PREOPS-3598/20230726T202836Z",
```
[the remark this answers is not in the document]
hsinfang: Somehow `black` moves this second part of the string to line up with `collections` ...  

But your suggestion looks better to my eyes too.

bin.src/make_latiss_export.py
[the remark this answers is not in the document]
kfindeisen: Sorry, I'm still confused. If `--target-repo` isn't set, does that mean the import overwrites the old files, or crashes?

bin.src/make_latiss_export.py
kfindeisen: If you're keeping the possibility of `target_butler` being `None`, can you add a comment mentioning that that's what the `AttributeError` catch is for? It's a bit unconventional.

bin.src/make_latiss_export.py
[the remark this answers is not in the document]
hsinfang: I will modify and just use `_filter_datasets` from `MiddlewareInterface` -- that's exactly the wanted thing here.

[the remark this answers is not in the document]
hsinfang: Hmm... I'm not sure what happens if `butler import` some conflicting datasets. I'll try it.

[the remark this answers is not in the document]
hsinfang: Just to see what'd happen, I `butler import` a conflicting `skyMap` dataset in my test repo, and got 
```
sqlalchemy.exc.IntegrityError: (sqlite3.IntegrityError) UNIQUE constraint failed: dataset_location.datastore_name, dataset_location.dataset_id
[SQL: INSERT INTO dataset_location (datastore_name, dataset_id) VALUES (?, ?)]
[parameters: [('FileDatastore@<butlerRoot>', '6e7fc9280dc34e1bae635fb63be46216'), ('FileDatastore@<butlerRoot>', 'f677ff862b1c40949dade39395052cd1')]]
``` 

Good to see that an error was generated.

bin.src/make_latiss_export.py
```diff
@@ -0,0 +1,110 @@
...
+    logging.basicConfig(level=logging.INFO, stream=sys.stdout)
+
+    args = _make_parser().parse_args()
+    src_butler = daf_butler.Butler(args.src_repo)
+
+    logging.info("Exporting datasets and collections")
+    start = time.time_ns()
+    _export_for_copy(src_butler)
+    end = time.time_ns()
+    logging.info("Export finished in %.3fs.", 1e-9 * (end - start))
```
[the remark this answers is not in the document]
hsinfang: Thanks for the suggestion. Indeed very handy!

bin.src/make_latiss_export.py
hsinfang: Instead of catching `target_butler` being `None` , it now makes a new, empty repo, so the end effect is the same but it looks clearer I think.

hsinfang:
@kfindeisen  may you please take another look?  Thanks to your feedbacks I think the script looks better now. It still assumes some peculiarities of `/repo/embargo` for those datasets and collections, but probably the script could be made useful relatively easily when we want to do something similar in the future.

bin.src/make_latiss_export.py
```diff
@@ -0,0 +1,108 @@
...
+    return parser
+
+
+def main():
+    logging.basicConfig(level=logging.INFO, stream=sys.stdout)
+
+    args = _make_parser().parse_args()
+    src_butler = daf_butler.Butler(args.src_repo)
+
+    with time_this(msg="Exporting datasets and collections", level=logging.INFO):
```
kfindeisen, line 74: Note that this message is printed _after_ the export is finished.

bin.src/make_latiss_export.py
```diff
@@ -43,6 +46,14 @@ def _make_parser():
         required=True,
         help="The location of the repository from which datasets are exported.",
     )
+    parser.add_argument(
+        "--target-repo",
+        required=False,
+        help="The location of the repository to which datasets are exported. "
+             "Datasets already existing in the target repo will not be "
+             "exported from the source repo, If no target repo is given, all "
```
kfindeisen, line 54:
```suggestion
             "exported from the source repo. If no target repo is given, all "
```

bin.src/make_latiss_export.py
```diff
@@ -52,40 +63,66 @@ def main():
     args = _make_parser().parse_args()
     src_butler = daf_butler.Butler(args.src_repo)
 
+    if args.target_repo:
+        target_butler = daf_butler.Butler(args.target_repo, writeable=False)
+    else:
+        # If no target_butler ig given, create an empty one.
```
kfindeisen, line 69:
```suggestion
        # If no target_butler is given, create an empty one.
```

bin.src/make_latiss_export.py
```diff
@@ -52,40 +63,66 @@ def main():
...
+        temp_dir = tempfile.TemporaryDirectory()
+        config = daf_butler.Butler.makeRepo(temp_dir.name)
+        target_butler = daf_butler.Butler(config)
+
     with time_this(msg="Exporting datasets and collections", level=logging.INFO):
-        _export_for_copy(src_butler)
+        _export_for_copy(src_butler, target_butler)
+
+    if not args.target_repo:
+        temp_dir.cleanup()
```
kfindeisen, line 78: I guess a context manager is impractical here, so please put this in a `finally` block.

bin.src/make_latiss_export.py
```diff
@@ -52,40 +63,66 @@ def main():
...
     """Export selected data to make copies in another butler repo.
 
     Parameters
     ----------
     butler : `lsst.daf.butler.Butler`
         The source Butler from which datasets are exported.
+    target_butler : `lsst.daf.butler.Butler`
+        The target Butler to which datasets are exported. If given,
+        it is checked to avoid exporting existing datasets. No
+        checks are done to verify if datasets are really identical.
```
kfindeisen, line 91: I think this description was written when `target_butler` was still optional.

kfindeisen (approved):
Looks good, thanks! Minor comments only.

bin.src/make_latiss_export.py
hsinfang: I couldn't come up with a way that makes the `finally` block looks better than the context manager, so I'm going with the context manager.

bin.src/make_latiss_export.py
hsinfang: Message corrected

bin.src/make_latiss_export.py
hsinfang: Corrected

bin.src/make_latiss_export.py
hsinfang: Because as-it it only saves the collection if the collection doesn't already exist (that'd mean all of these LATISS collections this time), and the `refcats` collection already exists in the dev central repo, I'd still edit the chain by hand this time instead of changing the script for it.

PR #85, opened 2023-09-25
DM-40163: Rework dataset transfer code in MiddlewareInterface

kfindeisen:
This PR removes the export/import idiom from MiddlewareInterface. The old idiom needed to use the deprecated `butler.datastore.root` to anchor relative paths, and there's no guarantee that the central repo won't have its datasets split among multiple stores.

### DIFF
--- a/python/activator/middleware_interface.py
+++ b/python/activator/middleware_interface.py
@@ -378,9 +378,11 @@ class MiddlewareInterface:
 
         with tempfile.NamedTemporaryFile(mode="w+b", suffix=".yaml") as export_file:
             with self.central_butler.export(filename=export_file.name, format="yaml") as export:
-                self._export_refcats(export, center, radius)
-                self._export_skymap_and_templates(export, center, detector, wcs, self.visit.filters)
-                self._export_calibs(export, self.visit.detector, self.visit.filters)
+                export.saveDatasets(self._export_refcats(center, radius))
+                export.saveDatasets(
+                    self._export_skymap_and_templates(center, detector, wcs, self.visit.filters))
+                export.saveDatasets(self._export_calibs(self.visit.detector, self.visit.filters),
+                                    elements=[])
                 self._export_collections(export, self._get_template_collection())
                 self._export_collections(export, self.instrument.makeUmbrellaCollectionName())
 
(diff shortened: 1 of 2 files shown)


python/activator/middleware_interface.py
```diff
@@ -868,19 +919,19 @@ def _export_subset(self, exposure_ids: set[int],
...
-            self.central_butler.import_(filename=export_file.name,
-                                        directory=self.butler.datastore.root,
-                                        skip_dimensions={"instrument", "detector",
-                                                         "skymap", "tract", "patch"},
-                                        transfer="copy")
+        # Transferring governor dimensions in parallel can cause deadlocks in
+        # central registry. We need to transfer our exposure/visit dimensions,
+        # so handle those manually.
+        for dimension in ["exposure", "visit"]:
+            for record in self.butler.registry.queryDimensionRecords(
```
timj, line 926: Does this work because we don't expect many exposure/visit records in the source butler? If there were a lot the alternative is to extract the relevant exposure/visit records directly from the refs that are being transferred.

timj (approved):
This looks good to me. Do you notice any performance difference? We know that export/import via YAML is really slow compared to doing the direct transfer.

python/activator/middleware_interface.py
timj: Right. I had missed that the where clause had a limited set of exposure IDs.

I don't really know what the data flow is here. Are you saying that the data turn up in prompt processing, the exposure record is first defined there, then you define the visits there, and then this is where all the results get copied to permanent storage? In that case it does seem like all the other visit dimensions will need to be included.

timj: (unless @TallJimbo tells me that something clever is going on inside syncDimensionData)

kfindeisen: bq. Are you saying that the data turn up in prompt processing, the exposure record is first defined there, then you define the visits there, and then this is where all the results get copied to permanent storage?

Yes, that's correct. The raw is first ingested into the local repo (since we're in a hurry). It's later separately ingested into the central repo, so we don't _strictly_ need to transfer the `exposure` dimension in production, but it's here because our testbeds can't emulate that step. Visit definition is, as far as I know, done _only_ here in the local repo.

kfindeisen: All three of the other `visit` dimensions have a [`populated_by: visit` line](https://github.com/lsst/daf_butler/blob/main/python/lsst/daf/butler/configs/dimensions.yaml#L430) in their configs, but AFAICT it's not actually _used_ anywhere in `daf_butler`.

TallJimbo: This does look exactly like the use case for [DM-34838](https://jira.lsstcorp.org/browse/DM-34838) which is about making those `populated_by` entries actually do something.  And that means this is indeed going to leave out all those auxiliary dimension records.  For now I think the list of dimension elements to export needs to be expanded to also include:

- `visit_definition`
- `visit_detector_region`
- `visit_system`
- `visit_system_membership`

And you're right that this has gone unnoticed because in the absence of those transfers it would just be future queries against the central repo that would be affected.

kfindeisen:
> Do you notice any performance difference? We know that export/import via YAML is really slow compared to doing the direct transfer.

Actually, it seems about the same: the integration test for this ticket had downloads of <2.4-3.8 s for `prep_butler`, and <6.9-9.0 s for `export_outputs`, while a previous run of `main` had <2.6-3.2 s and <8.2-9.0 s, respectively.

timj:
The speed up mainly comes when you have large numbers of datasets. Switching execution butler to transfer_from from export/import made a huge difference for 100k datasets.

PR #93, opened 2023-11-07
DM-40822: Update Prompt Processing docs to remove references to the IDF

kfindeisen:
This PR removes documentation and code that was valid only with the old Google Cloud environment. It also stops the vestigial Google Artifact Registry upload, which we haven't been using for either USDF deployments or `prompt-proto-service` builds.

This PR should not be merged before November 10 or before a `w_2023_45` base is available, because it will overwrite the current `w_2023_44` container (no behavior changes will occur, but it causes bookkeeping problems for the biweekly AuxTel run).


hsinfang (approved):
Looks good!

kfindeisen:
Merging one day early; I'll make sure any containers are built using the latest daily rather than weekly 44.

PR #94, opened 2023-11-09
DM-41658: Prompt Processing can't configure surveys with "-"

kfindeisen:
This PR makes the pipeline config parsing more flexible regarding survey names, now allowing any character that isn't clearly problematic.


kfindeisen:
Added one last-minute commit to log incoming `nextVist` messages -- it's something that would have been very useful last night!

hsinfang (approved):
All look great! Thanks for having these improvements so quickly!

PR #96, opened 2023-11-14
DM-41664: Rename prompt_prototype package and containers

kfindeisen:
This PR updates all code and docs from `prompt_processing` to `prompt_prototype`, and renames the containers from `prompt-proto-*` to `prompt-*`. The only remaining references to "proto" are those to `prompt-proto-service`, which will be renamed at a later date.


hsinfang (approved):
LGTM

PR #97, opened 2023-11-14
DM-41717: Prompt Processing observing support for AuxTel week 46

kfindeisen:
This is a running PR for quick fixes needed during observing.

PR #98, opened 2023-11-16
DM-41717: Prompt Processing observing support for AuxTel week 46

kfindeisen:
This is a running PR to test Prompt Processing fixes and the `BASE_TAG_LIST` system. It uses a different branch name from #97 to ensure that `DM-41717-d_2023_11_14` is not overwritten.

### DIFF
--- a/python/activator/visit.py
+++ b/python/activator/visit.py
@@ -54,9 +54,12 @@ class BareVisit:
         OPEN = 2
         EITHER = 3
 
-    salIndex: int               # this maps to an instrument
+    # script queue that generated the event. One queue usually runs one telescope, but they can switch
+    salIndex: int
     scriptSalIndex: int
-    groupId: str                # observatory-specific ID; not the same as visit number
+    # Observatory-specific ID. Same as Butler's group_name, not the same as
+    # Butler's group_id or visit number
+    groupId: str
     coordinateSystem: CoordSys  # coordinate system of position
     # (ra, dec) or (az, alt) in degrees. Use compare=False to exclude from hash.
     position: list[float] = field(compare=False)
@@ -104,11 +107,11 @@ class FannedOutVisit(BareVisit):
 @dataclass(frozen=True, kw_only=True)
 class SummitVisit(BareVisit):
     # Extra fields are in the NextVisit messages from the summit
-    private_efdStamp: float = 0.0
+    private_efdStamp: float = 0.0  # time of visit publication; UTC in unix seconds
     private_kafkaStamp: float = 0.0
     private_identity: str = "ScriptQueue"
     private_revCode: str = "c9aab3df"
     private_origin: int = 0
-    private_seqNum: int = 0
+    private_seqNum: int = 0        # counts script calls since queue start. Not the same as Butler seq_num
     private_rcvStamp: float = 0.0
-    private_sndStamp: float = 0.0
+    private_sndStamp: float = 0.0  # time of visit publication; TAI in unix seconds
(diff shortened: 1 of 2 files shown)


python/activator/visit.py
```diff
@@ -54,9 +54,12 @@ class Dome(enum.IntEnum):
         OPEN = 2
         EITHER = 3
 
-    salIndex: int               # this maps to an instrument
+    # script queue that generated the event. One queue usually runs one telescope, but they can switch
+    salIndex: int
     scriptSalIndex: int
-    groupId: str                # observatory-specific ID; not the same as visit number
+    # Observatory-specific ID. Same as Butler's group_name, not the same as
+    # Butler/RubinTV's group_id or visit number
```
hsinfang, line 61: I'm not sure we need to mention RubinTV because (1) RubinTV just shows whatever from Butler with a thin name mapping (2) RubinTV might change and that can be confusing to others reading the comment in the future.   It's actually easy to get RubinTV to call the column differently [(slack thread here);](https://lsstc.slack.com/archives/C04SW09ECCE/p1699981196663919) I was thinking to PR that, but then think that can add even more confusion...

python/activator/visit.py
```diff
@@ -104,11 +107,11 @@ def get_bare_visit(self):
...
+    private_efdStamp: float = 0.0  # time of visit publication; UTC in unix seconds
     private_kafkaStamp: float = 0.0
     private_identity: str = "ScriptQueue"
     private_revCode: str = "c9aab3df"
     private_origin: int = 0
-    private_seqNum: int = 0
+    private_seqNum: int = 0        # queue-specific counter. Not the same as Butler/RubinTV seq_num
     private_rcvStamp: float = 0.0
-    private_sndStamp: float = 0.0
+    private_sndStamp: float = 0.0  # time of visit publication; TAI in unix seconds
```
hsinfang, line 117: Trying to dig out again what these `private_*` fields mean and thinking whether adding more info would be useful. Just dumping them here

[Sasquatch](https://sasquatch.lsst.io/user-guide/timestamps.html)
> SAL timestamps are added automatically by SAL and have the private_ prefix. In particular, the private_sndStamp timestamp is the time at which the message was published by the CSC in TAI, which should be very close to the time at which the telemetry was measured. In fact, many telemetry topics rely on the private_sndStamp timestamp for that.
> The private_sndStamp timestamp is also converted to UTC and recorded in the private_efdStamp field, which is used as the time index in the EFD.


[slack thread](https://lsstc.slack.com/archives/CM65SGAJE/p1681916893293629)
> private_sndStamp, which should be the TAI instant when the SAL message was issued from the CSC, presumably only a short time after the actual event. 
> private_efdStamp should be the same value in UTC

And [slack thread](https://lsstc.slack.com/archives/CM65SGAJE/p1681918857875009?thread_ts=1681915811.873869&cid=CM65SGAJE)
> private_sndStamp ... is the time the sample was published.. this is TAI but is converted to UTC (and stored in private_efdStamp
> most topics will rely on private_sndStamp as also being the time the measurement was performed..

hsinfang: After reading them again, I think what you've put in the comments are perfect summary for prompt processing context.

hsinfang (approved):
Looks good! Those extra debug logs will be very useful.

python/activator/visit.py
kfindeisen: Yes, I used the Sasquatch page as my source.

Though given the example of `private_seqNum`, we should probably avoid using the `private_` fields if possible. 😬

kfindeisen:
(Oops, apparently you can review a PR that's "not ready for review"!)

PR #100, opened 2023-11-27
DM-34141: Distinguish retriable and non-retriable failures in Prompt Processing

kfindeisen:
This is a dummy PR to test the new error-handling code.


dspeck1 (approved):
Looks good.  Like the text in the mock side effects!

PR #102, opened 2023-12-05
DM-42027: new stack segfault with libpq

hsinfang:


kfindeisen:
Hmm... looks like there's a bug in the GHA config; it's building the base container but not the service one. 😞

hsinfang:
I thought it's a deliberate choice not to build a new service container with just the `Dockerfile.main` change?

Dockerfile.main
```diff
@@ -6,5 +6,6 @@ RUN source /opt/lsst/software/stack/loadLSST.bash \
         flask \
         gunicorn \
         python-confluent-kafka \
+        openssl=3.1.4 \
```
kfindeisen, line 9: Would it be possible to add a comment here that this is a workaround for DM-XXXXX?

I'm not familiar with mamba, so I don't know how this line will interact with Conda's `rubin-env`. I think it would be good to test this before merging.

kfindeisen (approved):
Change makes sense, but please do test on `dev`.

kfindeisen:
> I thought it's a deliberate choice not to build a new service container with just the `Dockerfile.main` change?

Sorry, just saw your comment on Jira. You're right, you'd have to override `BASE_TAG_LIST` for a service build to be meaningful anyway.

timj:
I think conda-forge have fixed the libpq problem now.

hsinfang:
As of earlier today when I tested `d_2023_12_05` on prompt processing `dev` it still crashed. It had openssl 3.2.0.  

Not familiar with our whole stack building process well enough, and not finding a jira DM ticket... how do I know if/when the workaround is no longer necessary, except deploying to `dev` to test? 

I'll definitely test in `dev` before touching `prod`.

PR #103, opened 2023-12-05
DM-41730: Exception logs don't have metadata

kfindeisen:
This PR augments `RecordFactoryContextAdapter` to log exceptions with the context in which the exception was raised, not the one in which it was caught.

### DIFF
--- a/tests/test_logger.py
+++ b/tests/test_logger.py
@@ -319,22 +319,61 @@ class AddLogContextTest(unittest.TestCase):
                     self.assertEqual(e.logging_context, {"color": "blue", "pid": 42})
                     # pid=42 should have been removed here
                     self.log.error("Exception caught")
+                    # Exception should hang on to old state
+                    self.log.exception("Exception detected")
+                    self.log.warning("Implied exception", exc_info=True)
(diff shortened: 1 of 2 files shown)


tests/test_logger.py
```diff
@@ -325,6 +326,16 @@ def test_exception_handling(self):
...
                               ])
 
+    def test_overwriting_exception_handling(self):
+        factory = logging.getLogRecordFactory()
+        try:
+            with factory.add_context(color="blue"):
+                with factory.add_context(color="red"):
+                    raise RuntimeError("Something failed")
+        except RuntimeError as e:
+            # Preserved original context while escaping multiple managers?
```
hsinfang: Do you mean to leave the "?" mark here?

hsinfang (approved):
Looks good to me. The unit test really helps understand what's going on!

tests/test_logger.py
kfindeisen: Sorry for the ambiguous grammar. Will fix.

kfindeisen:
> The unit test really helps understand what's going on!

Is there something that could be done to make the code clearer?

PR #105, opened 2023-12-06
Use custom configuration of MPGraphExecutor.

kfindeisen:
This is a dummy PR for testing whether our logging problems are caused by MPGraphExecutor.


kfindeisen:
Closing in favor of #106.

PR #106, opened 2023-12-07
DM-42063: No task logs in Prompt Processing output

kfindeisen:
This PR turns off multiprocessing in pipeline execution, preventing a bug where task logs were not going through our logger. This is a mid-term workaround; we expect to restore compatibility with multiprocessing in the future.

### DIFF
--- a/python/activator/middleware_interface.py
+++ b/python/activator/middleware_interface.py
@@ -35,6 +35,7 @@ import astropy
 
 from lsst.resources import ResourcePath
 import lsst.afw.cameraGeom
+import lsst.ctrl.mpexec
 from lsst.ctrl.mpexec import SeparablePipelineExecutor
 from lsst.daf.butler import Butler, CollectionType
 import lsst.dax.apdb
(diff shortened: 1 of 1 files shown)


python/activator/middleware_interface.py
```diff
@@ -776,6 +777,37 @@ def ingest_image(self, oid: str) -> None:
...
+        executor : `lsst.ctrl.mpexec.QuantumGraphExecutor`
+            The executor to use.
+        """
+        quantum_executor = SingleQuantumExecutor(
+            butler,
+            factory,
+        )
+        graph_executor = MPGraphExecutor(
+            # TODO: re-enable parallel execution once we can log as desired with CliLog or a successor
+            # (see issues linked from DM-42063)
```
hsinfang, line 804: Looks like a good workaround for the time being. Maybe we should create a new ticket for the future work discussed at Slack on more properly engineering this even if that may not happen soon?

hsinfang (approved):

python/activator/middleware_interface.py
kfindeisen: I think it will need to be done piece by piece, was hoping to discuss the details with Middleware after catching up on backlog.

PR #108, opened 2023-12-12
DM-41591: Include a visit with no template coverage in the upload.py test

hsinfang:


doc/playbook.rst
```diff
@@ -112,6 +119,8 @@ To inspect them with the MinIO Client ``mc`` tool, first set up an alias (e.g. `
 
 For Butler not to complain about the bucket names, set the environment variable ``LSST_DISABLE_BUCKET_VALIDATION=1``.
 
+See also `this note <https://confluence.lsstcorp.org/display/LSSTOps/USDF+S3+Bucket+Organization>`_ on USDF S3 Bucket Organization.
+
```
kfindeisen, line 123: Should this link be higher up (close to the list of Buckets)?

kfindeisen (commented):
I'm not sure I understand the Kafdrop instructions; can you clarify?

In the future, please put changes about different topics on separate commits.

doc/playbook.rst
```diff
@@ -101,8 +106,10 @@ The AWS CLI can be used to inspect non-tenenat buckets:
```
kfindeisen, line 106: Do these instructions need to be updated? I've been trying to get the file timestamps so I can write up an issue, and even with `AWS_PROFILE` and `LSST_DISABLE_BUCKET_VALIDATION` set:

```
$ s3 --profile rubin-prompt-processing ls s3://rubin-summit/

An error occurred (NoSuchBucket) when calling the ListObjectsV2 operation: Unknown
$ s3 --profile rubin-prompt-processing ls s3://rubin:rubin-summit/

Parameter validation failed:
Invalid bucket name "rubin:rubin-summit": Bucket name must match the regex "^[a-zA-Z0-9.\-_]{1,255}$" or be an ARN matching the regex "^arn:(aws).*:(s3|s3-object-lambda):[a-z\-0-9]*:[0-9]{12}:accesspoint[/:][a-zA-Z0-9\-.]{1,63}$|^arn:(aws).*:s3-outposts:[a-z\-0-9]+:[0-9]{12}:outpost[/:][a-zA-Z0-9\-]{1,63}[/:]accesspoint[/:][a-zA-Z0-9\-]{1,63}$"
```

kfindeisen: Ah, found the problem: to look up `rubin-summit`, you need to **not** use the developer profile, but the default LSST one. 😵‍💫

hsinfang: `LSST_DISABLE_BUCKET_VALIDATION` is our stack thing and not recognized by AWS CLI. I don't think we know a way to use AWS CLI with a tenant prefix.  I've been using `mc` for it. Hopefully the updates around line 109 in this PR clarifies it; otherwise I should rephrase better.

kfindeisen: Sorry, my bad for overlooking the `mc` docs.

doc/playbook.rst
```diff
@@ -78,9 +78,14 @@ The bucket ``rubin:rubin-pp-users`` holds:
 * ``rubin:rubin-pp-users/unobserved/`` contains raw files that the upload script(s) can draw from to create incoming raws.
 
 ``rubin:rubin-pp`` has had notifications configured for it; these publish to a Kafka topic.
+To check what messages are sent, do port-forward:
+.. code-block:: sh
+    kubectl port-forward svc/kafdrop -n kafka 9000:9000
+
+Then messages can be viewed locally via Kafdrop at `http://127.0.0.1:9000/topic/rubin-prompt-processing`_.
```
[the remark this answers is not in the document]
hsinfang: This would be one way to check the file arrival notifications from the bucket, not the nextVisit messages.  If I understand correctly, they are on separate kafkas (nextVisit from sasquatch's and prompt processing has its own).

I got the tip from [this slack thread ](https://lsstc.slack.com/archives/C0345PUU4N9/p1692740715825959?thread_ts=1692738210.807919&cid=C0345PUU4N9) when we were debugging some ceph setup.  I swear it was working at one point, probably from my laptop. 

However,  it seems no longer working probably because we didn't move the service to phalanx when we transitioned.  I'll check with Dan.   I'll leave out this part as it isn't really relevant to this ticket anyway.

hsinfang:
Funny thing is I originally had each topic on separate commits and squashed them thinking it was too many.  They are now separate again.

doc/playbook.rst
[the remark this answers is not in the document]
kfindeisen: My mistake with mixing up the two Kafkas. Unfortunately, the instructions on the Slack thread also don't make it clear which system you're supposed to do what on...

kfindeisen (approved):

PR #112, opened 2023-12-15
DM-41713: Many calibs (bias/flat/dark) preloaded in LATISS testing

kfindeisen:
This PR replaces our first call to `queryDatasetAssociations` with a much faster loop over individual datasets. We still need the _second_ call to `queryDatasetAssociations`, which provides the information we need to recertify the calibs we've preloaded (see [DM-41915](https://jira.lsstcorp.org/browse/DM-41915)), but that call will be skipped if calib preload becomes a no-op.

### DIFF
--- a/python/activator/middleware_interface.py
+++ b/python/activator/middleware_interface.py
@@ -1192,20 +1192,20 @@ def _filter_calibs_by_date(butler: Butler,
         guaranteed to be the same `~lsst.daf.butler.DatasetRef` objects passesd
         to ``unfiltered_calibs``, but guaranteed to be fully expanded.
     """
+    # Unfiltered_calibs can have up to one copy of each calib per certify cycle.
+    # Minimize redundant queries to find_dataset.
+    unique_ids = {(ref.datasetType, ref.dataId) for ref in unfiltered_calibs}
     t = Timespan.fromInstant(astropy.time.Time(date, scale='utc'))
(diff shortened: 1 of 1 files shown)


python/activator/middleware_interface.py
```diff
@@ -1190,26 +1190,20 @@ def _filter_calibs_by_date(butler: Butler,
...
-                _log_trace.debug("%s (valid over %s) does not match %s.", ref, associations[ref].timespan, t)
-        else:
-            _log_trace.debug("No calib associations for %s.", ref)
+    for dataset_type, data_id in unique_ids:
+        # Use find_dataset to simultaneously filter by validity and chain order
+        found_ref = butler.find_dataset(dataset_type,
+                                        data_id,
+                                        collections=collections,
+                                        timespan=t,
+                                        dimension_records=True,
```
TallJimbo, line 1205: I can't tell without more context, but it may be worth seeing if you can get away with setting this to `False`.  Calibrations shouldn't need any dimensions beyond what you get from registering an instrument.

TallJimbo (approved):
Looks good.  I also have an idea for the second pass you referred to in the PR description, but I'll put that on the Jira ticket you referenced.

python/activator/middleware_interface.py
kfindeisen: No, we need expanded refs for `transfer_from`. Worse, the requirement only causes problems in production (because it's triggered by ObsCore).

PR #114, opened 2024-01-12
DM-42469: Update testers and fan-out for the nextVisit schema change

hsinfang:


python/activator/visit.py
```diff
@@ -84,21 +85,19 @@ def __str__(self):
...
-    instrument: str             # short name
     detector: int
     private_sndStamp: float     # time of visit publication; TAI in unix seconds
 
     def __str__(self):
         """Return a short string that disambiguates the visit but does not
         include "metadata" fields.
         """
-        return f"(instrument={self.instrument}, groupId={self.groupId}, survey={self.survey} " \
+        return f"(groupId={self.groupId}, survey={self.survey} " \
```
kfindeisen, line 95: I think it's still useful to keep `instrument` as one of the "identifier" fields, even if the activator does almost immediately filter on it...

python/tester/utils.py
```diff
@@ -220,7 +220,7 @@ def send_next_visit(url, group, visit_infos):
                    f"filters: {info.filters} ra: {info.position[0]} dec: {info.position[1]} "
                    f"instrument: {info.instrument} survey: {info.survey}")
         records_level = dict(value=asdict(info))
-        value_schema_level = dict(value_schema_id=1, records=[records_level])
+        value_schema_level = dict(value_schema_id=97, records=[records_level])
```
kfindeisen, line 223: I'm a bit worried about this arbitrary-looking number. Will we have to update this every time something changes in the `dev` Kafka?

kfindeisen (approved):
I'm a bit worried about the schema ID, otherwise looks good!

python/tester/utils.py
hsinfang: It seems yes. I'm also worried about the not-very-well-documented update process related to Sasquatch redeployment and schema registry changes.

kfindeisen (approved):
Sorry, I missed your re-review request somehow. `startTime` looks good, and I think it reflects typical AuxTel observations.

python/tester/utils.py
hsinfang: We might alternatively send the messages with the full schema every time (a long string with all fields, their types. The `value_schema` on [Sasquatch doc](https://sasquatch.lsst.io/user-guide/restproxy.html)).  I'm not sure that'd be better in this case.

kfindeisen: Probably not, if that's not what the summit is doing.

hsinfang: I couldn't track down what summit does and will just continue to use schema id for now. I'll add some more docs in [DM-42618](https://jira.lsstcorp.org/browse/DM-42618)

PR #116, opened 2024-01-25
DM-42227: Switch from datetime to astropy.time in MiddlewareInterface

kfindeisen:
This PR removes the use of `datetime` in `MiddlewareInterface` in favor of `astropy.time`, though some unit tests still use `datetime` as the oracle. It also fixes a bug in our tests that was exposed by lsst/dax_apdb#41.

### DIFF
--- a/python/activator/middleware_interface.py
+++ b/python/activator/middleware_interface.py
@@ -523,9 +523,7 @@ class MiddlewareInterface:
         # supported in queryDatasets yet.
         calib_where = f"detector={detector_id} and physical_filter='{filter}'"
         # TAI observation start time should be used for calib validity range.
-        # private_sndStamp is the visit publication time in TAI, not UTC,
-        # but difference shouldn't matter.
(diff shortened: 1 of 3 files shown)


hsinfang (approved):

hsinfang:
A moment... Still checking if it makes sense to update
https://github.com/lsst-dm/prompt_processing/blob/main/python/tester/upload.py#L293

kfindeisen:
To be fully self-consistent, we probably should (if I understand things correctly, the translation between YMD and seconds since epoch is affected by the UTC-TAI offseet). Though it's worth noting that the real T&S code [uses a homebrew system](https://github.com/lsst-ts/ts_utils/blob/develop/python/lsst/ts/utils/tai.py#L142) on top of `datetime` instead of `astropy.time`.

hsinfang:
I confirmed that the `DATE-BEG` metadata from json or fitsheader is TAI (by checking numbers of one LATISS image  and [here](https://confluence.lsstcorp.org/display/SYSENG/Auxiliary+Telescope+Header+Information+Topic+Mapping ) because I don't know where is the offcial header document). 
`private_sndStamp` is in TAI too.

PR #117, opened 2024-01-29
DM-33453: Update pipeline references in response to RFC-775.

erinleighh:
Changed `HyperSuprimeCam` to `HSC` and `DarkEnergyCamera` to `DECam` in response to RFC-775.

### DIFF
--- a/pipelines/HSC/ApPipe.yaml
+++ b/pipelines/HSC/ApPipe.yaml
@@ -1,7 +1,7 @@
 description: End to end Alert Production pipeline specialized for HSC-RC2
 
 imports:
-  - location: $AP_PIPE_DIR/pipelines/HyperSuprimeCam/ApPipe.yaml
+  - location: $AP_PIPE_DIR/pipelines/HSC/ApPipe.yaml
     exclude:
       # TODO: container does not yet support R/B analysis.
       - rbClassify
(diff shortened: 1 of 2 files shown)


kfindeisen (approved):
Looks good! However, please don't merge this until at least the day after lsst/ap_pipe#150, so that we can get a daily container that uses these paths. It's a bit of unfortunate timing that this ticket got done just as we're starting a new AuxTel run...

Pinging @hsinfang so she's aware of the potential conflict.

kfindeisen:
Please also update `tests/data/ApPipe.yaml`. I missed it myself until I did a final search for references.

The good news is that file should break unit tests, so we might not have to worry about premature container builds. Though that's a very hacky solution to the problem. 😬

kfindeisen:
I've just restarted the service job based on [the `d_2024_01_31` container](https://github.com/lsst-dm/prompt_processing/pkgs/container/prompt-base/173296409?tag=d_2024_01_31). I'll run a quick HSC integration test, and then you should be good to go!

kfindeisen:
[A minimum HSC run of the full `ApPipe.yaml`](https://grafana.slac.stanford.edu/d/2ME8kLK4z/prompt-processing-development-output?from=1706723700000&to=1706724300000) finished without errors. Thanks for your patience!

PR #120, opened 2024-02-02
DM-41241: Create a new dev butler central repo in s3://rubin-pp-dev-users and move there

hsinfang:


doc/playbook.rst
```diff
@@ -155,13 +155,12 @@ To inspect buckets with the MinIO Client ``mc`` tool, first set up an alias (e.g
     mc alias set usdf-pp https://s3dfrgw.slac.stanford.edu ACCESS_KEY SECRET_KEY
```
kfindeisen, line 165: Should the instructions for AWS credentials also be updated to use a `prompt-processing-dev` profile? I find it less confusing if the profile name matches the user name.

doc/playbook.rst
```diff
@@ -310,7 +309,7 @@ Install the Prompt Processing code, and set it up before use:
     setup -r prompt_processing
 
 The tester scripts send ``next_visit`` events for each detector via Kafka on the ``next-visit-topic`` topic.
-They then upload a batch of files representing the snaps of the visit to the ``rubin:rubin-pp`` S3 bucket, simulating incoming raw images.
+They then upload a batch of files representing the snaps of the visit to the ``rubin-pp-dev`` S3 bucket, simulating incoming raw images.
 
 Eventually a set of parallel processes running on multiple nodes will be needed to upload the images sufficiently rapidly.
```
kfindeisen, line 324: We already have this! 😁 (Well, we didn't need the "multiple nodes" part.)

doc/playbook.rst
```diff
@@ -121,10 +121,20 @@ The bucket ``rubin-pp-dev-users`` holds:
 
 * ``rubin-pp-dev-users/unobserved/`` contains raw files that the upload script(s) can draw from to create incoming raws.
 
-``rubin-pp-dev`` has had notifications configured for it; these publish to a Kafka topic.
+``rubin-pp-dev`` has had notifications configured for it; these publish to the Kafka topic ``prompt-processing-dev``.
+The notifications can be viewed at `Kafdrop <https://k8s.slac.stanford.edu/usdf-prompt-processing-dev/kafdrop>`_.
+.
```
kfindeisen, line 126: Extra period. Should the Kafdrop link have been included in the previous commit?

doc/playbook.rst
```diff
@@ -121,10 +121,20 @@ The bucket ``rubin-pp-dev-users`` holds:
...
+
+The buckets ``rubin:rubin-pp`` and ``rubin:rubin-pp-users`` are also for Prompt Processing development and previously used by the testers.
```
kfindeisen, line 128: I think the "Buckets" section has grown long enough that it needs to be broken up into sections. Maybe "Current Buckets", "Legacy Buckets", and "AWS Credentials"?

doc/playbook.rst
```diff
@@ -121,10 +121,20 @@ The bucket ``rubin-pp-dev-users`` holds:
...
+``rubin:rubin-pp-users`` contains an older version of the development central repository.
+``rubin:rubin-pp`` has notifications configured to publish to the Kafka topic ``rubin-prompt-processing``.
+
+The Ceph user ``rubin-prompt-processing`` owns the ``rubin:rubin-pp`` and ``rubin:rubin-pp-users`` buckets, and the newer Ceph user ``prompt-processing-dev`` owns the ``rubin-pp-dev`` and ``rubin-pp-dev-users`` buckets.
+We are in the process to deprecate the ``rubin-prompt-processing`` user as it has a more restrictive permission than ``prompt-processing-dev``.
```
kfindeisen, line 133:
```suggestion
We are in the process of deprecating the ``rubin-prompt-processing`` user as it has more restrictive permissions than ``prompt-processing-dev``.
```

doc/playbook.rst
```diff
@@ -121,10 +121,20 @@ The bucket ``rubin-pp-dev-users`` holds:
...
 
 The default Rubin users' setup on ``rubin-devl`` includes an AWS credential file at the environment variable ``AWS_SHARED_CREDENTIALS_FILE`` and a default profile without read permission to the prompt processing buckets.
-A separate credential for prompt processing developers is at  `vault <https://vault.slac.stanford.edu/ui/vault/secrets/secret/show/rubin/usdf-prompt-processing-dev/s3-buckets>`_ and can be set up as another credential profile for Butler or command line tools such as AWS Command Line Interface and MinIO Client.
+A separate credential for prompt processing developers as the Ceph user ``prompt-processing-dev`` (new) or ``rubin-prompt-processing`` (old) is at  `vault <https://vault.slac.stanford.edu/ui/vault/secrets/secret/show/rubin/usdf-prompt-processing-dev/s3-buckets>`_.
```
kfindeisen, line 136: Mention that to get `rubin-prompt-processing` you need to look at an older version of the secret? (Or is there another way?)

doc/playbook.rst
```diff
@@ -145,7 +155,7 @@ The AWS CLI can be used to inspect non-tenenat buckets:
 
    You must pass the ``--endpoint-url`` argument even if you have ``S3_ENDPOINT_URL`` defined.
 
-Some of the prompt processing buckets are Ceph tenant buckets and require a tenant prefix, which violates the bucket name standard and is not supported by AWS CLI.
+If a bucket is a Ceph tenant bucket and requires a tenant prefix, the bucket name violates the standard and is not supported by AWS CLI.
```
kfindeisen, line 158: Say explicitly that these are the buckets that start with `rubin:`?

kfindeisen (approved):
Looks good! I'm concerned about the ObsCore config, but I assume it's already been applied.

etc/db_butler.yaml
```diff
@@ -1,5 +1,128 @@
...
+    datasets: lsst.daf.butler.registry.datasets.byDimensions.ByDimensionsDatasetRecordStorageManagerUUID
+    datastores: lsst.daf.butler.registry.bridge.monolithic.MonolithicDatastoreRegistryBridgeManager
+    dimensions: lsst.daf.butler.registry.dimensions.static.StaticDimensionRecordStorageManager
+    opaque: lsst.daf.butler.registry.opaque.ByNameOpaqueTableStorageManager
+    obscore:
+      cls: lsst.daf.butler.registry.obscore._manager.ObsCoreLiveTableManager
+      config:
+        # Configuration copied from /repo/embargo
+        namespace: embargo
+        version: 1
```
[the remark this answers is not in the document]
hsinfang: I don't understand what's going on behind the scenes, but the `s3://rubin-pp-dev-users/central_repo/butler.yaml` file does not have this long config block, just like `s3://rubin-summit-users/butler.yaml`  not having the long config block. 

I'm worried about hidden drift between `/repo/embargo` and our dev repo too, but don't have a better idea than keeping that in mind and following what's happening with `/repo/embargo` .

bin.src/make_remote_butler.py
```diff
@@ -53,11 +53,13 @@ def _make_parser():
                         help="The config file to use for the new repository. Defaults to etc/db_butler.yaml.")
     parser.add_argument("--export-file", default="export.yaml",
                         help="The export file containing the repository contents. Defaults to ./export.yaml.")
+    parser.add_argument("--instrument", default="HSC",
+                        help="The short name of the instrument.")
```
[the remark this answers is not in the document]
hsinfang: Indeed. This time I used it with HSC data to start, but it doesn't need to be that way. I removed the default value.

kfindeisen:
Oops! Sorry, I forgot one request: could [your instructions for running the pipeline remotely](https://jira.lsstcorp.org/browse/DM-41241?focusedCommentId=4658999#comment-4658999) to update the output dataset types be added to the playbook? We'll probably need to add more types in the future, and it would be helpful not to have to figure out the right collections, data IDs, etc.

doc/playbook.rst
hsinfang: 👍  . Update this in a [separate commit ](https://github.com/lsst-dm/prompt_processing/pull/120/commits/a50fe44385d7e0003b506af0db16d0c891c71d03)

doc/playbook.rst
hsinfang: IIRC the "previous commit" was the commit changing all old bucket names to the new ones. As the Kafdrop setup is independent of the bucket change, to me it looks clearer to keep one commit purely about the bucket change, and another commit about updating the playbook.

doc/playbook.rst
hsinfang: Great idea! The section is reorganized and rewritten in https://github.com/lsst-dm/prompt_processing/pull/120/commits/3d07119c17885cb8213e5bd7e998a99f23df9edf

doc/playbook.rst
kfindeisen: Ok, I thought it was the result of setting up the bucket.

PR #122, opened 2024-02-12
DM-41154: Do not skip missing files silently in Butler transfers

kfindeisen:
This PR adds a manual integrity check to the transfers to and from the central repo. If any dataset could not be transferred, the service will now emit a warning (if it can be worked around) or error (if it cannot). (If _excess_ datasets are transferred, it will also warn, though the diagnostics won't be helpful.)


timj (approved):
Looks reasonable to me.

PR #126, opened 2024-02-20
DM-42618: Add startTime to the nextVisit schema in prompt processing

hsinfang:


python/tester/upload.py
```diff
@@ -217,7 +218,7 @@ def get_samples_non_lsst(bucket, instrument):
             dome=FannedOutVisit.Dome.OPEN,
             duration=float(EXPOSURE_INTERVAL+SLEW_INTERVAL),
             totalCheckpoints=1,
-            private_sndStamp=hsc_metadata[exp_id]["time"],
+            private_sndStamp=hsc_metadata[exp_id]["time"]-60,
```
kfindeisen, line 221: For self-consistency, it might be better to calculate the offset from `EXPOSURE_INTERVAL` and `SLEW_INTERVAL`.

python/tester/upload_hsc_rc2.py
```diff
@@ -190,7 +191,7 @@ def prepare_one_visit(kafka_url, group_id, butler, visit_id):
             dome=SummitVisit.Dome.OPEN,
             duration=float(EXPOSURE_INTERVAL+SLEW_INTERVAL),
             totalCheckpoints=1,
-            private_sndStamp=data_id.records["exposure"].timespan.begin.unix_tai,
+            private_sndStamp=data_id.records["exposure"].timespan.begin.unix_tai-60,
```
kfindeisen, line 194: This script also defines `EXPOSURE_INTERVAL` and `SLEW_INTERVAL`.

python/tester/utils.py
```diff
@@ -220,7 +220,7 @@ def send_next_visit(url, group, visit_infos):
                    f"filters: {info.filters} ra: {info.position[0]} dec: {info.position[1]} "
                    f"instrument: {info.instrument} survey: {info.survey}")
         records_level = dict(value=asdict(info))
-        value_schema_level = dict(value_schema_id=97, records=[records_level])
+        value_schema_level = dict(value_schema_id=99, records=[records_level])
```
kfindeisen, line 223: Given your later instructions for updating it, I'd recommend moving this to a global constant so that it's easy to find.

kfindeisen (approved):

python/tester/upload.py
hsinfang: Good idea.  I'm setting them to 2*duration before the exposure start time.  All these timestamps were in the past, as from the original raws' metadata, so they will need to be better simulated to be useful.

PR #128, opened 2024-02-28
Add compatibility guidelines to release process

kfindeisen:
This PR defines the API with respect to which we define compatibility for semantic versioning, and therefore what version number we should use for any given change. Since part of the API is the Butler `dimensions-config` version, it also adds the version(s) to the standard release description.


doc/playbook.rst
```diff
@@ -93,12 +112,15 @@ Add text as follows.
 * Any specific motivation for the release (for example, including a specific feature, preparing for a specific observing run)
 * Science Pipelines version and rubin-env version
 * Any changes to the APDB and Alerts schemas
+* Required `Butler dimensions-config`_ version(s)
+
+.. _Butler dimensions-config: https://github.com/lsst/daf_butler/blob/main/python/lsst/daf/butler/configs/dimensions.yaml
```
hsinfang, line 117: To be clear, is it correct that we want to note this version number from `daf_butler` as in the base `lsst_distrib` stack for the release?

hsinfang (approved):
Looks good and as what the team discussed and agreed on yesterday.

doc/playbook.rst
kfindeisen: The `dimensions-config` version of the central repo usually lags behind the base stack for compatibility reasons. I think we'd want to actually think about what we're supporting.

The current Prompt Processing codebase definitely works with version 3 or later, which was the version at which we became aware that it matters (DM-42038, #104). We have some export code that could be removed if we required version 4 or later, but we never bothered.

Going the other way, we would probably work with version 2 (https://github.com/lsst/daf_butler/commit/111f45eceda3b88d7c46216fb0355018f3ba7a1a shouldn't have effected us any more than https://github.com/lsst/daf_butler/commit/45a68664cbdc9fe8f2a5a04005eb86673fe44c02 did), but since there are no longer such repos I don't see much point.

dspeck1 (approved):
Looks good and covers what I am aware of.

doc/playbook.rst
kfindeisen: For the record, our dev repo is on version 4, and `/repo/embargo` is on version 5.

kfindeisen: @hsinfang, is there some way I could clarify that the supported versions are a matter of developer judgement? (Though linking to lsst/daf_butler#953's new changelog instead of the config file should help.)

timj: Butler repositories can only be migrated monotonically and at some point `Butler.transfer_from` between wildly different universes is going to fail. When we migrate /repo/embargo to version 6 you are likely going to want to migrate your repos (which you are motivated to do anyhow in this specific case). It depends on the specific change in the version (v5 was changing a size of a column so is irrelevant).

kfindeisen: `transfer_from` is not a problem because (per #104) the local repo is always initialized with the same version as the central repo. But yes, I expect that we'll make a major release in which we require support for `group`.

hsinfang: I'm thinking when we make a new release, what info about the butler dimension version needs to go to the release notes. If we are making a new release right now with a `w_2024_08` base stack, does it suffice just to say `w_2024_08` is on version 5?  Maybe yes because that implies some compatibilities with repos of whatever versions?

timj: The problem is that the current weekly default universe version doesn't have to have any relationship with the one in the butler repositories you are using. When I release version 6 it might be weeks before `/repo/embargo` gets migrated and that might be at a different time to when `/repo/main` is converted.

kfindeisen: I think we would need to say something like versions 3-5 -- no later, because support for 6 isn't in `w_2024_08`, but no earlier, because Prompt Processing makes assumptions about Butler features (for example, once we migrate to 6 we will almost certainly have references to the `group` dimension in Prompt Processing code, and may even write group-level datasets directly).

If we gave just the latest version, wouldn't we basically be saying that all repos (including `/repo/embargo`, which affects a lot of people) _must_ be upgraded immediately even if we don't actually need it?

I retconned [our 1.0.0 release description](https://github.com/lsst-dm/prompt_processing/releases/tag/v1.0.0) as an example of what I think it could look like (will probably change once we add APDB versions, which didn't exist in mid-December).

PR #131, opened 2024-03-02
DM-42726: Implement upload.py LSSTComCamSim for dev test

hsinfang:


etc/export_latiss.yaml
```diff
@@ -0,0 +1,23 @@
...
+- collections: refcats*
+  datasetType: atlas_refcat2_20220201
+- collections: refcats*
+  datasetType: gaia_dr3_20230707
+- collections: skymaps
+  datasetType: skyMap
+# Workaround: use a matching expression rather than a specific
+# string "LATISS/calib" for the collection argument, so to avoid
+# MissingCollectionError when the collection does not exist in
+# the target repo.
```
kfindeisen, line 13: Why do you need a workaround? `_filter_datasets` is supposed to be runnable on an empty destination repo...

bin.src/make_export.py
kfindeisen, line 1: I guess I'd worry a little about whether any variations in `make_hsc_rc2_export` and `make_template_export` were captured, but great idea!

etc/export_comCamSim.yaml
```diff
@@ -0,0 +1,19 @@
...
+# Workaround: use a matching expression rather than a specific
+# collection string for the collection argument, so to avoid
+# MissingCollectionError when the collection does not exist in
+# the target repo.
+# These are the non-raw collections in "u/jchiang/bfk_50240117_w_2024_04"
+- collections: "*u/jchiang/bfk_50240117_w_2024_04/20240131T061636Z"
+- collections: "*u/jchiang/ptc_50240117_w_2024_04/20240131T055006Z"
+- collections: "*u/jchiang/flat_50240117_w_2024_04/20240129T161508Z"
+- collections: "*u/jchiang/dark_50240117_w_2024_04/20240128T225538Z"
+- collections: "*u/jchiang/bias_50240117_w_2024_04/20240128T225209Z"
```
kfindeisen: As I understand it, this is just the first iteration of the sim. Is the plan to replace these later?

etc/export_comCamSim.yaml
kfindeisen, line 1: I suggest merging this commit with "Add export config file".

python/activator/middleware_interface.py
```diff
@@ -532,7 +532,8 @@ def _export_refcats(self, center, radius):
         # collection, so we have to specify a list here. Replace this
         # with another solution ASAP.
         possible_refcats = ["gaia", "panstarrs", "gaia_dr2_20200414", "ps1_pv3_3pi_20170110",
-                            "atlas_refcat2_20220201", "gaia_dr3_20230707"]
+                            "atlas_refcat2_20220201", "gaia_dr3_20230707",
+                            "uw_stars_20240130", "uw_stars_20240228"]
```
kfindeisen, line 536: Do we actually need the `0130` version if we have access to `0228`?

python/tester/utils.py
```diff
@@ -175,31 +176,37 @@ def make_hsc_id(group_id, snap):
     return exposure_id, {"EXP-ID": f"HSCE{exposure_id:08d}"}
 
 
-def make_latiss_id(group_id, snap):
-    """Generate an exposure ID that the Butler can parse as a valid LATISS ID.
+def make_lsst_id(group_id, snap, abbrev):
+    """Generate an exposure ID that the Butler can parse as a valid LSST ID.
```
kfindeisen, line 179: My intent when I first wrote `make_hsc_id` was that each instrument would have its own specialization. Is the `abbrev` really the only thing that will need to change for real LsstComCam or LsstCam?

python/tester/upload.py
```diff
@@ -56,12 +56,14 @@
 class Instrument:
     n_snaps: int
     n_detectors: int
+    sal_index: int = 0
```
kfindeisen, line 59: Good idea, given that (for better or worse) we're already assuming this in fan-out service.

python/tester/upload.py
```diff
@@ -56,12 +56,14 @@
...
 
 
 INSTRUMENTS = {
-    "LSSTCam": Instrument(2, 189 + 8 + 8),
-    "LSSTComCam": Instrument(2, 9),
-    "LATISS": Instrument(1, 1),
+    "LSSTCam": Instrument(2, 189 + 8 + 8, 1),
+    "LSSTComCam": Instrument(2, 9, 3),
```
kfindeisen, line 64: Isn't it more likely that the real ComCam will be on queue 1, given that that's the main telescope?

python/tester/upload.py
```diff
@@ -305,7 +305,8 @@ def get_samples_lsst(bucket, instrument):
             detector=_DETECTOR_FROM_RS[instrument][m["raft_sensor"]],
             groupId=md["GROUPID"],
             nimages=INSTRUMENTS[instrument].n_snaps,
-            filters=md["FILTBAND"],
+            # ComCam currently sets the FILTBAND header to null.
+            filters=md["FILTBAND"] or md["FILTER"],
```
kfindeisen, line 309: I assume this doesn't cause problems with our existing LATISS data?

doc/playbook.rst
```diff
@@ -389,6 +390,7 @@ This script draws images stored in the ``rubin-pp-dev-users`` bucket.
   One of the files, the unobserved group `2023-10-11T01:45:47.810`, has modified RA at a location with no templates.
   Astrometry is also expected to fail in WCS fitting.
   This visit can test pipeline fallback features.
+* For LSSTComCamSim, 1 group, in total 9 raw fits files and their corresponding json metadata files, are curated.
```
kfindeisen, line 393: This isn't enough to let us test visit coordination or source association. Will we have more groups once the simulations mature?

python/tester/upload.py
```diff
@@ -153,6 +153,38 @@ def main():
...
+def _add_to_raw_pool(raw_pool, snap_num, visit, blob):
+    """Add a detector-snap to the raw pool for uploading.
+
+    Parameters
+    ----------
+    raw_pool : mapping [`str`, mapping [`int`, mapping [`activator.FannedOutVisit`, `s3.ObjectSummary`]]]
+        Available raws as a mapping from group IDs to a mapping of snap ID.
+        The value of the innermost mapping is the observation metadata for
+        each detector, and a Blob representing the image taken in that
+        detector-snap.
```
kfindeisen, line 167: We really ought to get a proper class for this! 😂 Maybe I'll find an opportunity...

kfindeisen (approved):
Looks good! Mostly just clarification requests.

kfindeisen:
Oh, one question: does this PR make #132 obsolete?

doc/playbook.rst
hsinfang: We already have more groups. I'll add another set to `unobserved`.

python/tester/upload.py
hsinfang: Since our existing LATISS data have proper `FILTBAND`, the rest is just ignored.

python/tester/upload.py
hsinfang: Yes. Thank you for spotting it!

etc/export_comCamSim.yaml
hsinfang: Yes, that's my understanding too that this is only the first iteration and new set will be generated.  I assume only the final set will be copied into `/repo/embargo` where our prod deployment will use.  I don't think we really need to update the calibs in our dev central repo, because we aren't vetting the pipelines quality in dev but just ensure processing runs.   But, if we have doubts about prod for any reasons, we can update the dev repo too.

python/activator/middleware_interface.py
hsinfang: The `0130` version is what we currently have in the dev central repo.  And `0228` would be what we expect to use if we wanted `prod` today.   
I can copy the `0228` version into the dev central repo, and then `0130` wouldn't be needed here in this line.  But I hesitate to keep updating the dev repo just to get the newest data.

python/tester/utils.py
hsinfang: Actually `LsstComCam` and `LsstCam` share the same `abbrev` .   ¯\\_(ツ)_/¯  
I assume it's from the camera DAQ and the sim runs on the real ComCam DAQ?

Indeed, so far it looks to me that all LSST cameras (LATISS/LSSTComCam/LSSTComCamSim/LSSTCam) can share this same `make_lsst_id` in mocking an exposure metadata for the upload test.

bin.src/make_remote_butler.py
```diff
@@ -97,46 +96,6 @@ def _add_chains(butler, instrument_name):
...
-    # Chain calibration collections
-    instrument = Instrument.fromName("HSC", butler.registry)
-    butler.registry.setCollectionChain(
-        instrument.makeCalibrationCollectionName(),
-        [
-            "HSC/calib/DM-32378",
-            "HSC/calib/gen2/20180117",
-            "HSC/calib/DM-28636",
-        ],
-    )
```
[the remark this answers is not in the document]
hsinfang: This fix up wasn't really needed in the past, because we can export the `HSC/calib` chain from `/repo/main`.   IIRC I was just too confused back then, or I couldn't figure out the correct parameters for `flattenChains` and `includeChains`, and using a fix-up seemed easier back then.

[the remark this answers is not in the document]
hsinfang: Also, at the time we didn't have the (HSC/)templates collection set up in `/repo/main`, so some kind of fix-up was necessary back then anyway.

hsinfang:
Yes. https://github.com/lsst-dm/prompt_processing/pull/132 will be discarded.

python/activator/middleware_interface.py
kfindeisen: That makes sense, but is `uw_stars_20240228` actually deployed (or expected to be deployed) in `/repo/embargo`?

bin.src/make_latiss_export.py
```diff
@@ -86,65 +96,36 @@ def _export_for_copy(butler, target_butler):
...
-                contents.saveCollection(collection)
+        if "collections" in wants:
+            for selection in wants["collections"]:
+                for collection in butler.registry.queryCollections(**selection):
+                    logging.debug(f"Selecting collection {collection}")
+                    try:
+                        target_butler.registry.queryCollections(collection)
+                    except daf_butler.registry.MissingCollectionError:
+                        # MissingCollectionError is raised if the collection does not exist in target_butler.
+                        contents.saveCollection(collection)
```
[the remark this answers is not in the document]
hsinfang: I thought that updating the collection manually might have been safer, but actually I can re-import the new chained collection.  I just tested `butler import` with an export yaml including a chain that already exists in the target repo but the chain is defined differently in the export yaml; it just overwrote the chain without errors or warnings.   So, yes, I think we can update a collection using the export yaml made by this script.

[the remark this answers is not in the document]
kfindeisen: Ok, that's unexpected; I thought `import` was really strict about everything matching.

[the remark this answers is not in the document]
hsinfang: I didn't look into the source code but I guess this is similar to `butler query-collections` which would just redefine the chain and overwrite without warnings.

etc/export_latiss.yaml
hsinfang: it doesn't if I constrain it for a specific collection and that collection doesn't exist in the destination repo.  It's an extra keyword arguments to [`dest_butler_repo.registry.queryDatasets(*args, **kwargs)`](https://github.com/lsst-dm/prompt_processing/blob/main/python/activator/middleware_interface.py#L1238)  and 
`queryDatasets(...,collections="something-not-in-repo")` gives `MissingCollectionError`.  

I guess usually we only use it with well-defined collection names and those names are registered in the empty local butler repo before any `_filter_datasets`?

kfindeisen: Yes, but we could have a handler similar to https://github.com/lsst-dm/prompt_processing/blob/main/python/activator/middleware_interface.py#L1239-L1242. Can you ticket it as a bug with a how-to-reproduce? It should be an easy fix.

hsinfang: That'd be a great addition to `_filter_datasets` --> https://rubinobs.atlassian.net/browse/DM-43294

python/activator/middleware_interface.py
hsinfang: I don't know what will be added to `/reop/embargo` for the actual ops rehearsal. I'll remove `uw_stars_20240228` and wait to add the eventual refcat when that's known.

hsinfang: On second thought, if the current `obs_lsst` default is `uw_stars_20240228`, copying the new refcat to pp's dev repo is probably better than a config override.  I'll check more.

bin.src/make_export.py
hsinfang: Honestly I'm a little worried about that too.  This probably should have been done before [DM-41241](https://rubinobs.atlassian.net/browse/DM-41241) and we could have felt more reassured.

bin.src/make_latiss_export.py
```diff
@@ -86,65 +96,36 @@ def _export_for_copy(butler, target_butler):
         The target Butler to which datasets are exported. It is checked
         to avoid exporting existing datasets. No checks are done to
         verify if datasets are really identical.
+    wants : `dict` of `list` of `dict`
+        A `dict` with optional keys "datasets" and "collections" containing
+        a list of selection expressions to identify what datasets and
+        collections are to be selected. The selection expression has the
+        keyworded argument dictionary which is passed to butler to query
+        datasets or collections.
```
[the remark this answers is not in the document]
hsinfang: Thank you for pointing out that session in the dev guide -- really useful to reformat it: https://github.com/lsst-dm/prompt_processing/pull/131/commits/39fe81e1ddf19bc27d41b78d32c84308b4c9ed01

python/activator/middleware_interface.py
hsinfang: Change of mind.... 
`uw_stars_20240228` has now been imported to the dev repo. I changed the commit and use only `uw_stars_20240228`.

PR #132, opened 2024-03-05
DM-42712: Add support for raws from the new LsstComCamSim camera

hsinfang:


hsinfang:
The full set of changed related to LSSTComCamSim support is at https://github.com/lsst-dm/prompt_processing/pull/131

PR #134, opened 2024-03-06
DM-43182: Prompt Processing GitHub Actions don't account for test-only PRs

kfindeisen:
This PR modifies the Prompt Processing build actions to trigger when the unit tests are updated, to confirm that the new tests pass. It also places the "Release CI" workflow under the same restrictions as "Build and test service", so that it does not trigger on non-code changes.

Unfortunately, since both workflows trigger on changes to the workflow, it's not possible to test these changes on this PR.


dspeck1 (approved):
LGTM

PR #137, opened 2024-03-14
Tickets DM-40415: Add doPackageAlerts and doProduceAlerts to ApPipe.yaml

bsmartradio:


hsinfang (approved):
LGTM.  I think the two commits can be squashed together. Or if you want them separately, the commit messages can be clearer on what each does.

kfindeisen:
I'm confused... don't we want alerts for ComCamSim and _not_ HSC or LATISS?

hsinfang:
Form the PR I thought maybe @bsmartradio  wants to turn it on for LATISS first and decide about LSSTComCamSim later?

bsmartradio:
> I'm confused... don't we want alerts for ComCamSim and _not_ HSC?

Yes, I assumed this was for the Latiss testing. I may have misunderstood your slack comment earlier. Should I be adding to ComCamSim to ensure its part of the ops test, even though we haven't tested that yet? If we are just concerned about Latiss testing, I can leave out the HSC yaml.

bsmartradio:
@hsinfang , since the code freeze is happening tomorrow, I added producer config to LSSTComCamSim so that it can make it in for the ops rehearsal. I removed the HSC one because that one is of less concern and we probably want to keep it simple right now with everything else going on.

PR #139, opened 2024-03-20
DM-43394: Set up AP pipeline variant that omits forced-source loading

kfindeisen:
This PR adds a temporary pipeline that turns off forced source loading (and, with future changes to the pipeline, possibly forced source measurement as well). This is a configuration we may need for the Ops Rehearsal, but it is not expected to become the baseline `ApPipe`.

Because the config flag that is set here will not be available until `w_2024_12`, I have not tested the pipeline on our dev cluster. I have, however, confirmed locally that the pipeline definition is valid.

I've also piggybacked a timing improvement that may help us characterize the runtime problems that inspired the `doLoadForcedSources` config in the first place.

### DIFF
--- a/pipelines/LATISS/ApPipe-noForced.yaml
+++ b/pipelines/LATISS/ApPipe-noForced.yaml
@@ -1,0 +1,15 @@
+description: >-
+  Alert Production pipeline specialized for LATISS,
+  with forced source handling turned off
+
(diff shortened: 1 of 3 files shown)


kfindeisen:
With the expanded scope, this PR should be merged after lsst/ap_association#198.

ebellm (approved):

PR #141, opened 2024-03-22
DM-43445: Camera load in Prompt Processing assumes a single unbounded collection

kfindeisen:
This PR modifies the code for loading the initial `camera` dataset to get it from `<instrument>/calib` where possible, keeping the old behavior of `<instrument>/calib/unbounded` only as a fallback for older repos.

### DIFF
--- a/python/activator/middleware_interface.py
+++ b/python/activator/middleware_interface.py
@@ -270,13 +270,21 @@ class MiddlewareInterface:
         self._init_ingester()
         self._init_visit_definer()
 
-        # HACK: explicit collection gets around the fact that we don't have any
-        # timestamp/exposure information in a form we can pass to the Butler.
-        # This code will break once cameras start being versioned.
-        self.camera = self.central_butler.get(
-            "camera", instrument=self.instrument.getName(),
-            collections=self.instrument.makeUnboundedCalibrationRunName()
-        )
+        # TODO: can replace this with find_dataset on DM-42825
+        try:
+            self.camera = self.central_butler.get(
+                "camera",
+                instrument=self.instrument.getName(),
+                collections=self.instrument.makeCalibrationCollectionName(),
+            )
+        except lsst.daf.butler.DatasetNotFoundError:
+            # Repos that don't allow retrieval of camera from <instrument>/calibs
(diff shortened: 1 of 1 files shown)


python/activator/middleware_interface.py
```diff
@@ -270,13 +270,20 @@ def __init__(self, central_butler: Butler, image_bucket: str, visit: FannedOutVi
...
+                collections=self.instrument.makeCalibrationCollectionName(),
+            )
+        except lsst.daf.butler.DatasetNotFoundError:
+            # Repos that don't allow retrieval of camera from <instrument>/calibs
+            # have <instrument>/calibs/unbounded.
+            self.camera = self.central_butler.get(
+                "camera",
+                instrument=self.instrument.getName(),
+                collections=self.instrument.makeUnboundedCalibrationRunName(),
+            )
```
TallJimbo, line 286: I think this is a possible alternative:
```
camera_ref = self.central_butler.find_dataset(
    "camera",
    instrument=self.instrument.getName(),
    collections=self.instrument.makeCalibrationCollectionName(),
    timespan=Timespan(begin=None, end=None)
)
self.camera = butler.get(camera_ref)
```
it's a shame `Butler.get` doesn't accept `timespan` and forward that down to its own `find_dataset` call directly.

Your original looks fine, but if what I've written works, it'll avoid dependence on one of our shakier convention.

TallJimbo (approved):

python/activator/middleware_interface.py
timj: I will look at forwarding timespan in [DM-43499](https://rubinobs.atlassian.net/browse/DM-43499).

kfindeisen: I'd like to give it a shot, but the "live" tests have turned up serious problems that will keep me busy up until the code freeze. I'll add a TODO comment and try your suggestion as part of [DM-42825](https://rubinobs.atlassian.net/issues/DM-42825), which already involves expanding our use of `find_dataset`.

timj: This support was added in lsst/daf_butler#982

PR #142, opened 2024-03-22
DM-42739: Configure Prompt Processing production service for Sasquatch upload

kfindeisen:
This PR allows upload of `analysis_tools` metrics to a configurable namespace.

### DIFF
--- a/python/activator/middleware_interface.py
+++ b/python/activator/middleware_interface.py
@@ -1148,6 +1148,18 @@ class MiddlewareInterface:
             else:
                 _log.warning("No datasets match visit=%s and exposures=%s.", self.visit, exposure_ids)
 
+        # TODO: can we use SasquatchDatastore to streamline this?
+        dispatcher = _get_sasquatch_dispatcher()
+        if dispatcher:
+            with lsst.utils.timer.time_this(_log, msg="upload metrics", level=logging.DEBUG):
+                # Making bundles a collection makes debug log simpler, and it should be short.
+                bundles = list(self._query_datasets_by_storage_class(self.butler, exposure_ids, output_runs,
+                                                                     "MetricMeasurementBundle"))
+                for bundle in bundles:
+                    _log_trace.debug("Uploading %s...", bundle)
+                    dispatcher.dispatchRef(self.butler.get(bundle), bundle)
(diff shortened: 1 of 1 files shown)


python/activator/middleware_interface.py
```diff
@@ -1125,27 +1126,40 @@ def export_outputs(self, exposure_ids: set[int]) -> None:
...
+
+        finally:
+            # TODO: can we use SasquatchDatastore to streamline this?
+            dispatcher = _get_sasquatch_dispatcher()
+            if dispatcher:
+                with lsst.utils.timer.time_this(_log, msg="upload metrics", level=logging.DEBUG):
+                    bundles = list(self._query_datasets_by_storage_class(
+                        self.butler, exposure_ids, output_runs, "MetricMeasurementBundle"))
+                    for bundle in bundles:
+                        _log_trace.debug(f"Uploading {bundle}...")
```
isullivan, line 1160: I thought f-strings were too inefficient inside loops for debugging logs (since they are evaluated even if the log is not written).

python/activator/middleware_interface.py
```diff
@@ -1261,6 +1275,46 @@ def _chain_exports(self, output_chain: str, output_runs: collections.abc.Iterabl
...
+
+        Yields
+        ------
+        dataset : `lsst.daf.butler.DatasetRef`
+            A dataset in ``collections`` of type ``storage_class``. Guaranteed
+            to include values for implied dimensions, but need not include
+            dimension records. The order in which datasets are returned
+            is undefined.
+        """
+        matching_types = {dtype for dtype in butler.registry.queryDatasetTypes(...)
```
isullivan, line 1301: Would there be any value in using an ordered dict here instead?

isullivan (approved):
A couple very minor comments.

python/activator/middleware_interface.py
kfindeisen: Yes, sorry, I got sloppy.

python/activator/middleware_interface.py
kfindeisen: I don't see how. I assume the dataset types would be the keys, but what would the values be?

isullivan: Sorry, I incorrectly read the {.. as defining a dict, and I was looking for ways to make the order more consistent. It is not important.

PR #149, opened 2024-03-29
DM-43584: Add an option to turn off caching entirely

kfindeisen:
This PR adds an emergency option that turns off dataset caching entirely, for situations where the local repo must not be allowed to grow at all.

Closes #148.

### DIFF
--- a/python/activator/middleware_interface.py
+++ b/python/activator/middleware_interface.py
@@ -64,6 +64,8 @@ _log_trace3.setLevel(logging.CRITICAL)  # Turn off by default.
 # VALIDITY-HACK: local-only calib collections break if calibs are not requested
 # in chronological order. Turn off for large development runs.
 cache_calibs = bool(int(os.environ.get("DEBUG_CACHE_CALIBS", '1')))
+# TODO: workaround for DM-40193
+cache_anything = bool(int(os.environ.get("DEBUG_CACHE_INPUTS", '1')))
 
 
 def get_central_butler(central_repo: str, instrument_class: str):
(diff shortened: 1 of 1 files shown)


python/activator/middleware_interface.py
```diff
@@ -64,6 +64,8 @@
 # VALIDITY-HACK: local-only calib collections break if calibs are not requested
 # in chronological order. Turn off for large development runs.
 cache_calibs = bool(int(os.environ.get("DEBUG_CACHE_CALIBS", '1')))
+# TODO: workaround for DM-40193
+cache_anything = bool(int(os.environ.get("DEBUG_CACHE_INPUTS", '1')))
```
ktlim, line 68: The transitions from Helm values `true`/`false` to K8s YAML `1`/`0` to Python `True`/`False` are a little mind-bending, but I guess it's better than getting `true` and `True` confused.

python/activator/middleware_interface.py
```diff
@@ -1370,6 +1372,11 @@ def clean_local_repo(self, exposure_ids: set[int]) -> None:
                     self.butler.registry.removeCollection(member)
                 self.butler.removeRuns(calib_runs, unstore=True)
 
+            # TODO: workaround for DM-40193
```
ktlim, line 1375: If this moved up to just after the `refresh()`, couldn't you put a `return` after it and skip all the other work, while still being nearly as compact code-wise?  But I guess it isn't really that much extra work, and in fact getting rid of other dataset types might speed up the `...` query.

ktlim (approved):

python/activator/middleware_interface.py
kfindeisen: Unfortunately, since environment variables must be strings, and any nonempty string casts to `bool` as `True`, there needs to be an explicit translation step _somewhere_. I picked this convention mainly because it mirrors the existing `LSST_DISABLE_BUCKET_VALIDATION`.

python/activator/middleware_interface.py
kfindeisen: Nope, doesn't work -- I don't know if it's extra registry entries or something else, but the rearranged order leaves  about 190 MB in the pods instead of 2 MB. I'm worried that may grow in time.

PR #150, opened 2024-03-29
Update Release Management docs

kfindeisen:
This PR improves the release management guidelines in the Playbook in three ways:
- It provides links to look up the latest versions or version histories of dependencies, so that we're aware of what versions are relevant to us (e.g., there is not yet an APDB schema 0.2, but we'd want to know if there is).
- It clarifies how to handle changes internal to Science Pipelines (just writing up what @hsinfang and I have already been doing).
- ~It removes most references to the alert schema version, as I learned from @bsmartradio that this is set dynamically and is not a function of the Science Pipelines release.~ It replaces old language about "alert schema changes" with references to the alert schema versioning policy.

### DIFF
--- a/doc/playbook.rst
+++ b/doc/playbook.rst
@@ -89,8 +89,8 @@ For the ``prompt_processing`` service, a new major version is triggered by any o
 
 * Incompatibility with old fanned-out ``nextVisit`` messages (almost any change to ``Visit`` qualifies)
 * Incompatibility with an old `APDB schema`_, `ApdbSql`_, or `ApdbCassandra`_ version (see `DMTN-269`_ for the distinction)
-* Breaking changes in the Alerts schema
 * Incompatibility with an old `Butler dimensions-config`_ version
+* A new major version of the `Alerts schema`_ (see `DMTN-093`_ for details)
 
 For the `next_visit_fan_out`_ service, a new major version is triggered by any of the following:
 
@@ -112,8 +112,10 @@ Add text as follows.
 * Any specific motivation for the release (for example, including a specific feature, preparing for a specific observing run)
 * Science Pipelines version and rubin-env version
 * Supported `APDB schema`_ and `ApdbSql`_/`ApdbCassandra`_ versions (see `DMTN-269`_ for rationale)
-* Any changes to the Alerts schema
 * Supported `Butler dimensions-config`_ versions
+* The `Alerts schema`_ version used for output (see `DMTN-093`_ for details)
+
+.. _DMTN-093: https://dmtn-093.lsst.io/#alertmanagement
 
 .. _DMTN-269: https://dmtn-269.lsst.io/
 
@@ -124,6 +126,8 @@ Add text as follows.
 .. _ApdbSql: https://github.com/lsst/dax_apdb/blob/main/python/lsst/dax/apdb/apdbSql.py#L60-L63
 
 .. _ApdbCassandra: https://github.com/lsst/dax_apdb/blob/main/python/lsst/dax/apdb/apdbCassandra.py#L80-L83
+
+.. _Alerts schema: https://github.com/lsst/alert_packet/blob/main/python/lsst/alert/packet/schema/latest.txt
 
 Select **Publish Release**.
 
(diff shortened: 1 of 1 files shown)


ebellm:
I'm confused by the statement from @bsmartradio that alert schema versions are not a function of Science Pipelines release.  The `alert_packet` package absolutely has a [schema version](https://github.com/lsst/alert_packet/blob/main/python/lsst/alert/packet/schema/latest.txt) which is connected to the pipelines release, and things will break if they are not versioned correctly.  So I think alert schema changes do need to trigger PP version changes.

Bri, were you thinking of the version numbers supplied by the Confluent Schema Registry?  I think we have to make those consistent with `alert_packet` rather than the other way around--https://rubinobs.atlassian.net/browse/DM-42606.

bsmartradio:
> I'm confused by the statement from @bsmartradio that alert schema versions are not a function of Science Pipelines release. The `alert_packet` package absolutely has a [schema version](https://github.com/lsst/alert_packet/blob/main/python/lsst/alert/packet/schema/latest.txt) which is connected to the pipelines release, and things will break if they are not versioned correctly. So I think alert schema changes do need to trigger PP version changes.
> 
> Bri, were you thinking of the version numbers supplied by the Confluent Schema Registry? I think we have to make those consistent with `alert_packet` rather than the other way around--https://rubinobs.atlassian.net/browse/DM-42606.

I may have misunderstood what Krzysztof meant by independent. I explained that the version of the alert schema that is used in the pipeline is pulled from alert_packet​ using `alertPack.get_uri_to_latest_schema()`, so it will pull the latest schema in from alert_packet, and that when the pipeline checks which schema to use it will default to the latest version which is set in the `schemaFile​` config.

kfindeisen:
More precisely, I'd understood Brianna's to mean that it's pulled from the `latest.txt` on GitHub, and not the one internal to the Stack release. Is that wrong?

ebellm:
It's definitely using the stack-internal `latest.txt`, not phoning out to Github.

kfindeisen:
Then maybe I'll ask again what I asked in DM. As of [Prompt Processing 2.5](https://github.com/lsst-dm/prompt_processing/releases/tag/2.5.0), we're using `w_2024_14`, which means [our internal `latest.txt` reads "6.1"](https://github.com/lsst/alert_packet/blob/w.2024.14/python/lsst/alert/packet/schema/latest.txt).

What does that actually mean from a compatibility standpoint? Why should anybody care that PP 2.5 has Alerts 6.1? For example, are we restricted to uploading alerts to version 6.x servers? Or is it other people who need to be able to handle 6.x alerts?

ebellm:
At minimum, I don't think it's possible to change the version of the alert schema without changing the science pipelines version.  More generally, the alert schemas have close ties to the APDB schema and are likely to break things if they are not versioned together.

kfindeisen:
That wasn't my question; we are already reporting the science pipelines version and the APDB schema versions, and this whole discussion assumes that the alert schema version follows from our choice of science pipelines version (otherwise, it is not something we could put in the release notes).

Why would somebody using the Prompt Processing release, or looking at its outputs, need to know what the Alerts version is? For example, the APDB schema version matters because if Prompt Processing tries to connect to an APDB running an unsupported version, it will simply fail to process any data.

ebellm:
Thanks for the clarification.  Consumers of the alert stream need to know the schema version in order to be able to decode the alerts, as we are not planning to ship alerts with their Avro schema.

kfindeisen:
Thanks. Just to make sure we're on the same page, would consumers need to upgrade their code if we moved from 6.1 to 7.0?

ebellm:
Yes, they would for a major version bump--see https://dmtn-093.lsst.io/#alertmanagement

kfindeisen:
@ebellm, can you take another look? In addition to rewriting the alerts section, I've added a sentence about the possible use of weekly release branches. See https://github.com/lsst-dm/prompt_processing/releases/tag/2.5.0 for possible boilerplate about the alerts version.

ebellm (approved):

ebellm:
Looks good.

PR #151, opened 2024-04-02
DM-43414: Update playbook to use apdb-cli in place of make_apdb.py

andy-slac:


doc/playbook.rst
```diff
@@ -249,7 +249,7 @@ One raw was ingested, visit-defined, and kept in the development central repo, s
 
 .. code-block:: sh
 
-   make_apdb.py -c db_url="sqlite:///apdb.db"
+   apdb-cli create-sql "sqlite:///apdb.db" apdb_config.py
    pipetask run -b s3://rubin-pp-dev-users/central_repo -i LATISS/raw/all,LATISS/defaults,LATISS/templates -o u/username/collection  -d "detector=0 and instrument='LATISS' and exposure=2023082900500 and visit_system=0" -p $PROMPT_PROCESSING_DIR/pipelines/LATISS/ApPipe.yaml -c diaPipe:apdb.db_url=sqlite:///apdb.db --register-dataset-types
```
kfindeisen, line 253: Hmm, this is a problem -- until DM-43416, `pipetask run` doesn't let us actually use `apdb_config.py`.

```suggestion
   pipetask run -b s3://rubin-pp-dev-users/central_repo -i LATISS/raw/all,LATISS/defaults,LATISS/templates -o u/username/collection  -d "detector=0 and instrument='LATISS' and exposure=2023082900500 and visit_system=0" -p $PROMPT_PROCESSING_DIR/pipelines/LATISS/ApPipe.yaml -c diaPipe:apdb.db_url=sqlite:///apdb.db --register-dataset-types

.. TODO: update pipetask call after DM-43416
```

doc/playbook.rst
```diff
@@ -470,17 +470,17 @@ To restore the APDB to a clean state, run the following:
 .. code-block:: sh
 
    psql -h usdf-prompt-processing-dev.slac.stanford.edu lsst-devl rubin -c 'drop schema "pp_apdb_latiss" cascade;'
-   make_apdb.py -c namespace="pp_apdb_latiss" \
-       -c db_url="postgresql://rubin@usdf-prompt-processing-dev.slac.stanford.edu/lsst-devl"
+   apdb-cli create-sql --namespace="pp_apdb_latiss" \
+       "postgresql://rubin@usdf-prompt-processing-dev.slac.stanford.edu/lsst-devl" apdb_config_latiss.py
```
kfindeisen, line 474: Replace the original `psql` command with the `--drop` flag?

kfindeisen (approved):
Looks good, one question about the reset operation.

doc/playbook.rst
andy-slac: Well, it's not a problem if you just ignore that file for now and keep using the same pipetask options.

doc/playbook.rst
andy-slac: There is a subtle difference - `--drop` only removes tables that it thinks should exist (based on current configuration). IF configuration changed beween two `create-sql` executions, then some tables may not be removed. `DROP SCHEMA` is cleaner in that respect.

kfindeisen: Yes, I did notice that, but I'm assuming new versions are more likely to add tables than remove them. 😀 Calling "drop schema" was mostly because we needed a  way to ensure compatibility "by hand".

PR #155, opened 2024-04-26
DM-43416: Add compatibility patch for APDB configuration

kfindeisen:
The Prompt Processing code assumes that the pipeline is configured directly through `diaPipe.apdb`, but lsst/ap_pipe#182 makes the default be `apdb_config_url`. This patch lets Prompt Processing keep using the old system until it can support the new one.

The unit test failure is expected, since we don't have a container that includes lsst/ap_pipe#182.

It may not be necessary to actually merge this patch, if our `latest` base container is never one that contains DM-43416 but not DM-44096,


isullivan (approved):
Looks good.

kfindeisen:
Failure is expected -- see PR description.

PR #157, opened 2024-05-01
DM-44096: Migrate Prompt Processing to external APDB configs

kfindeisen:
This PR replaces all internal references to `apdb.db_url` with `parameters:apdb_config`, and updates the service configuration interface accordingly. Closes #155.


hsinfang (approved):
LGTM. 

The diaPipe:apdb.db_url  config in the pipetask example to create new dataset types probably should be fixed too.

PR #158, opened 2024-05-02
DM-43949: Require Butler dimensions-config 6+ in Prompt Processing

kfindeisen:
This PR makes a number of improvements to Prompt Processing that were blocked on the need to support central repos with dimensions-config 3-5. In particular, code now makes explicit use of the `group` dimension.


hsinfang (approved):
LGTM. 

Just to be consistent, I think we should update the paths of the unobserved HSC raws for their group IDs, basically something like  [this](https://github.com/lsst-dm/prompt_processing/commit/fc7b345feb2420d30b75a263e9c6d4d73feeaf3d) but with their correct/new group IDs.

kfindeisen:
Thanks, I forgot about those paths. Do you mean that slot should be filled by the original HSC exposure ID (since HSC doesn't use groups), or that they should use the same "compressed date" form generated by the new code?

hsinfang:
Probably the original exposure IDs, as the “original” group IDs if ingested by butler? 

Though they don’t really matter anyway because they aren’t used, it’s just for consistency of the path pattern.

PR #159, opened 2024-05-08
Clarify how minor versions work for APDB.

kfindeisen:
Minor version updates to the APDB or Apdb implementations are documented as "backwards compatible", but this is used to mean that new clients can use old databases, not the other way around. This commit adds examples to the versioning guidelines to make it clear that we should not be claiming compatibility with future APDB versions except at the patch level.

### DIFF
--- a/doc/playbook.rst
+++ b/doc/playbook.rst
@@ -115,7 +115,8 @@ Add text as follows.
 
 * Any specific motivation for the release (for example, including a specific feature, preparing for a specific observing run)
 * Science Pipelines version and rubin-env version
-* Supported `APDB schema`_ and `ApdbSql`_/`ApdbCassandra`_ versions (see `DMTN-269`_ for rationale)
+* Supported `APDB schema`_ and `ApdbSql`_/`ApdbCassandra`_ versions (see `DMTN-269`_ for rationale).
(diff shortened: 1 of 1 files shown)


hsinfang (approved):
Thanks for adding the clarification!

doc/playbook.rst
```diff
@@ -115,7 +115,8 @@ Add text as follows.
 
 * Any specific motivation for the release (for example, including a specific feature, preparing for a specific observing run)
 * Science Pipelines version and rubin-env version
-* Supported `APDB schema`_ and `ApdbSql`_/`ApdbCassandra`_ versions (see `DMTN-269`_ for rationale)
+* Supported `APDB schema`_ and `ApdbSql`_/`ApdbCassandra`_ versions (see `DMTN-269`_ for rationale).
+  A stack running a given minor version is compatible with *older* APDBs but not necessarily newer ones; for example, a release whose baseline is APDB schema 1.4.0 can access a schema 1.0.0 or 1.4.1 database, but not schema 1.5.
```
andy-slac, line 119: Maybe say something like "compatible with *older* APDBs with the same major version, but..."?

andy-slac (approved):
Looks good.

PR #160, opened 2024-05-08
DM-44189: Remove ccdVisitId from Prompt Processing

kfindeisen:
This PR uses `Apdb.containsVisitDetector` for error handling in pipeline execution, replacing the old dimension packer-based hack.


kfindeisen:
Breaking change because this commit requires `d_2024_05_07` or later, which introduces other incompatibilities.

parejkoj (approved):
Looks good! I'm a little bit surprised that this is the only place that needs a change, but I guess that's a sign of a well designed system!

PR #161, opened 2024-05-09
DM-43948: Create region/time dataset in Prompt Processing

kfindeisen:
This PR writes a `RegionTimeInfo` dataset directly to the preload run so that it can be used by preprocessing tasks.


kfindeisen:
`d_2024_05_01` does not yet produce RegionTimeInfo from the pipeline, so it can't be used to confirm that the two sources don't interact.

hsinfang (approved):

PR #166, opened 2024-05-16
DM-44393: Prompt Processing does not transfer visit definitions

kfindeisen:
This PR fixes a bug introduced on #158, which broke the mapping between exposure and visit and caused many datasets to effectively "disappear" from the central repo, depending on the query used.


timj (approved):
Looks great.

PR #169, opened 2024-06-11
DM-40193: Disk space leak in prompt processing

kfindeisen:
This PR adds a dataset cache to ensure the local repo never exceeds a configured size. The implementation is quite messy and may need to be replaced later, but the basic configuration hooks should be permanent.

### DIFF
--- a/python/activator/evictingSet.py
+++ b/python/activator/evictingSet.py
@@ -1,0 +1,160 @@
+# This file is part of prompt_processing.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+
+__all__ = ["EvictingSet"]
+
+
+import abc
+import collections.abc
+from typing import Any, Self
+
+
+# TODO: implementing the full MutableSet API is overkill, and both the API's
+# reliance on updating operators and the default implementations in MutableSet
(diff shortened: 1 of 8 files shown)


python/activator/evictingSet.py
```diff
@@ -0,0 +1,365 @@
...
+
+import abc
+import collections.abc
+import copy
+from typing import Any, Self
+
+import numpy
+
+
+class EvictingSet(collections.abc.MutableSet):
```
dhirving, line 34: I'm not sure it's beneficial for this to derive from the `MutableSet` ABC.  I wonder if it might be simpler for this to HAVE a `set`, instead of BEING a `set`.

Some reasons I say that are:
1. There seem to be several methods implemented to conform with the set interface, but which aren't actually needed by the code using this class.  One of the big ideas of wrapping mutable state in a class is that it limits the scope of code you need to look at in order to understand everything about the state.  Extra unused code increases the burden on future maintainers, since they may not realize immediately that they can ignore it.
2. It seems to be causing problems with accidentally losing the type when interacting with other sets.
3. It makes it harder for future maintainers to "jump to definition" when trying to track problems down -- instead of just clicking "jump" they have to know that the ABC maps the `-` operator to `__sub__` which then gets provided as a mixin based on the other available methods etc.  And there may be some surprising interactions with the eviction logic and the mixins.

One possible way this interface could be structured without the ABC might be:
```
class EvictingSet:
    def add(refs: Iterable[DatasetRef]) -> None
    def contents() -> set[DatasetRef]
    def trim_to_size(max_size: int) -> set[DatasetRef] # Returns evicted items
    def track_access(ref: DatasetRef) -> None
    def copy() -> Self
```
which I think wouldn't change the consumer code much other than requiring `.contents()` to be called a few places, and lets you dump a lot of baggage associated with the ABC.

Or possibly the "tracking" and "eviction" parts of this could be two separate things... there could be a `UsageTracker` that just keeps tracks of how often datasets have been used and can summarize that, and then implement different eviction strategies as plain functions that get run at the end when you go to do the eviction.

kfindeisen: You're right, I was blinded by thinking of a cache as being, conceptually, a map (as I write this, I'm still having trouble thinking of it as anything else). I can try to reimplement (yet another extra day 😭), though I don't think I'd trim the API down quite as much as you propose -- I'd want at least single-insert and removal operations for future use.

However, I don't think "tracking" and "eviction" can be separated -- _what_ information needs to be stored (counts, capped counts, frequencies, timestamps, etc.) is specific to a given replacement algorithm, and I don't want to commit to one yet.

dhirving: Yeah that makes sense.

I'd say definitely don't reimplement right now, just something to think about when you go to do the more advanced eviction algorithms later.

erinleighh (approved):
Everything looked good, only one minor comment on a comment.

PR #170, opened 2024-06-13
DM-42924: Document use of apdb-cli metadata.

kfindeisen:
This PR adds an alternative way to check an APDB's metadata; it works for both PostgreSQL and Cassandra, but requires bucket credentials.


isullivan (approved):
Looks good

PR #171, opened 2024-06-13
DM-44816: Add documentation for Cassandra APDB

isullivan:


doc/playbook.rst
```diff
@@ -504,11 +507,67 @@ To restore the APDB to a clean state, run ``apdb-cli create-sql`` with ``--drop`
...
+
+   apdb-cli create-cassandra sdfk8sk001 sdfk8sk004 KEYSPACE APDB_CONFIG.py --replication-factor=3 --enable-replica
+
+Here ``sdfk8sk001`` and ``sdfk8sk004`` are two nodes within the Prompt Processing allocation, which are the ``contact_points`` used for the initial connection.
+All of the available nodes will be used.
+If direct access to the APDB from the RSP is required, note that the ``APDB_CONFIG.py`` will need to be modified by hand to fully specify the hosts:
+
+.. code-block:: python
+
+   config.contact_points=['sdfk8sk001.sdf.slac.stanford.edu', 'sdfk8sk004.sdf.slac.stanford.edu']
```
kfindeisen, line 561: This is incorrect. The config must be set up with `sdfk8sk001.sdf.slac.stanford.edu` in order to be accessible from the Prompt Processing service (why would you want to access it from the RSP?).

Under no circumstances should people be making personal tweaks to the config file; the current architecture is designed on the assumption that there is *one* standardized file used to access any APDB instance. Rather than doing anything clever, 

    apdb-cli create-cassandra sdfk8sk001.sdf.slac.stanford.edu sdfk8sk004.sdf.slac.stanford.edu \
        KEYSPACE APDB_CONFIG.py --replication-factor=3 --enable-replica

creates a file that is guaranteed valid and safe to use.

kfindeisen: I propose using the conventional keyspace (e.g., `pp_apdb_latiss`) and filename (`pp_apdb_latiss-dev.py`) so the reader doesn't have to hunt for it (the filename in the Postgres example doesn't match what we're actually using, so that should be fixed).

kfindeisen (changes requested):
Please fix the documentation about the node names; the current text does not work as written and encourages unsafe practices.

doc/playbook.rst
isullivan: That's interesting, I didn't realize `sdfk8sk001.sdf.slac.stanford.edu` would be necessary for Prompt Processing. The example I gave worked fine in my batch mode testing. I will remove the reference to the RSP and update the examples.

isullivan:
I changed the documentation about the node names, and reorganized the sections and subsections. Please take another look.

doc/playbook.rst
```diff
@@ -461,6 +461,9 @@ The server is visible from ``rubin-devl``, and can be accessed through, e.g.,
 
    psql -h usdf-prompt-processing-dev.slac.stanford.edu lsst-devl rubin
 
+Credentials
+-----------
+
 For passwordless login, create a ``~/.pgpass`` file with contents:
```
kfindeisen, line 467: Should there be a "Postgres" header here?

doc/playbook.rst
```diff
@@ -483,9 +515,30 @@ From ``rubin-devl``, new APDB schemas can be created in the usual way:
...
+All of the available nodes will be used.
+In the above example, ``pp_apdb_latiss`` is the Cassandra keyspace (similar to schema for Postgres), and ``pp_apdb_latiss-dev.py`` is the usual APDB config.
+
 Resetting the APDB
 ------------------
 
+Postgres
+^^^^^^^^
+
 To restore the APDB to a clean state, run ``apdb-cli create-sql`` with ``--drop`` option which recreates all tables:
```
kfindeisen, line 542: I'm not sure splitting up this section makes sense -- it's literally just the one argument.

kfindeisen (approved):
Much better, thanks!

doc/playbook.rst
isullivan: I was unsure of the best choice, since the Postgres credentials include those for the Butler, which are needed even if you are using Cassandra. I could separate out the one line that is for the Postgres APDB, but that seems awkward to split apart the `~/.pgpass` contents. I could add a sentence to the Cassandra APDB section noting that the pgpass file is still needed for the Butler.

doc/playbook.rst
isullivan: I agree, but since we have code examples it seemed more clear to just include both. I could remove the subsections and change the text to:

> To restore the APDB to a clean state, add the ``--drop`` option to  ``apdb-cli create-sql`` or ``apdb-cli create-cassandra``  which will recreate all tables:

doc/playbook.rst
kfindeisen: That's not actually true -- Butler can take credentials from either source (and in DM-44141 we moved ours to db-auth).

isullivan: OK, I added the Postgres header and added a comment in the Cassandra section about Butler credentials.

PR #173, opened 2024-06-15
DM-44822: Extend playbook with examples on new CLI commands

andy-slac:


doc/playbook.rst
```diff
@@ -558,6 +558,26 @@ To restore the APDB to a clean state, add the ``--drop`` option to  ``apdb-cli c
...
+   apdb-cli list-cassandra sdfk8sk001.sdf.slac.stanford.edu
+
+To delete an existing APDB instance in Cassandra cluster:
+
+.. code-block:: sh
+
+   apdb-cli delete-cassandra sdfk8sk001.sdf.slac.stanford.edu <keyspace>
+
+The command will ask for confirmation before it tries to delete specified keyspace.
+Cassandra creates keyspace snapshot before deleting it, so disk space is not released immediately, special administration tools have to be used to remove snapshots.
```
isullivan, line 579: Could you add a statement directing people where to ask (e.g. a Slack channel) to have snapshots permanently removed?

isullivan (approved):

doc/playbook.rst
andy-slac: Hmm, we do not have a dedicated channel for APDB business. I think we may be better to make snapshot cleaning a part of the regular management ops so that people do not need to worry about those snapshots. They take some space, but that should not be an issue, and we'll monitor disk space in any case.

doc/playbook.rst
```diff
@@ -558,6 +558,27 @@ To restore the APDB to a clean state, add the ``--drop`` option to  ``apdb-cli c
    apdb-cli create-sql --drop --namespace="pp_apdb_lsstcomcamsim" \
        "postgresql://rubin@usdf-prompt-processing-dev.slac.stanford.edu/lsst-devl" apdb_config_lsstcomcamsim.py
 
+Removing APDB Instances
+-----------------------
```
kfindeisen, line 562: I realize I'm butting in, but what exactly is the motivation here? In the dev environment we should only have one APDB per instrument, and while we may want to _reset_ it, we almost never want to just outright delete it -- doing so would break the dev service(s).

The only time we've tried to delete an APDB, permanently, were the ones created for sanity-checking the 1.0.0 and 1.1.0 migrations, and the versioning system seems to be a little more stable now.

andy-slac: My guess is that people will create multiple APDB instances for various tests. Postgres `usdf-prompt-processing-dev/lsst-devl` contains more than 100 schemas/namespaces for APDB instances. If people can create those instances freely, it may be worth adding an ability to drop them as well?

kfindeisen: Those namespaces are from the precursor processing group -- completely different workflow, and they don't (and shouldn't) read these docs.

andy-slac: So, should I drop all of that? I'm OK with whatever you propose, the help for `apdb-cli` should be sufficient in case someone needs it.

isullivan: Yes, let's drop this from the playbook documentation. The proper place would be in https://pipelines.lsst.io , and I will make a ticket to update the documentation there with the text you wrote for this ticket.

andy-slac: OK, closing this PR.

PR #175, opened 2024-06-26
DM-45022: Allow a `latest` base container from a branch 

hsinfang:


kfindeisen:
The diff looks fine as a branch change, but are you planning to merge this? I'd assumed it would go on `u/hfc/gha` or `u/hfc/patch` (since those changes _did_ affect the base containers you built). Is there a reason that wouldn't work?

kfindeisen (approved):

hsinfang:
You are right! I don't need to merge this.

PR #176, opened 2024-06-27
DM-43648: Old repos left behind on Prompt Processing worker reboot

kfindeisen:
This PR adds a startup check that removes any prior local repos (from a killed worker) that may be taking up space. This prevents worker failures from cascading to later requests.

If the local repo cannot be removed, the preferred solution is to abort startup entirely; killing and replacing the pod is slow, but guarantees a clean slate.

### DIFF
--- a/python/activator/middleware_interface.py
+++ b/python/activator/middleware_interface.py
@@ -1462,7 +1462,14 @@ class MiddlewareInterface:
             # Transferring governor dimensions in parallel can cause deadlocks in
             # central registry. We need to transfer our exposure/visit dimensions,
             # so handle those manually.
-            self._export_exposure_dimensions(exposure_ids)
+            self._export_exposure_dimensions(
+                self.butler,
+                self.central_butler,
+                where="exposure in (exposure_ids)",
+                bind={"exposure_ids": exposure_ids},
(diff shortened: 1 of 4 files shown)


python/activator/middleware_interface.py
```diff
@@ -103,6 +105,48 @@ def get_central_butler(central_repo: str, instrument_class: str):
...
+
+    Parameters
+    ----------
+    repo_dir : `str`
+        An absolute path to a local repo that is *not* being managed by
+        TemporaryDirectory or similar.
+    central_butler : `lsst.daf.butler.Butler`
+        Butler repo to which to attempt to flush the local repo's contents.
+    """
+    try:
```
hsinfang, line 123: Looks good for today.   This reads somewhat similar to `_export_subset` and I wonder if they can be refactored for better code reuse later.

hsinfang (approved):
Looks good and will help with our ephemeral storage issues

python/activator/middleware_interface.py
kfindeisen: Not easily. `_export_subset` tries to make its query as restrictive as possible, both to minimize side effects and to keep things fast. Here we don't have the context to do that. 😞

PR #181, opened 2024-07-15
DM-44605: Prompt processing can do ISR for LATISS calibrations 

hsinfang:


hsinfang:
@kfindeisen  do you mind taking a look of some additional commits for this ticket?

python/activator/middleware_interface.py
```diff
@@ -327,12 +321,15 @@ def __init__(self, central_butler: Butler, image_bucket: str, visit: FannedOutVi
...
                  prefix: str = "s3://"):
         self.visit = visit
+        # Usually prompt processing only cares about on-sky images and expects all of them
+        # to have equatorial coordinates and sky rotation angles. However here we only warn
+        # and not raise so to allow special cases such as using calibration frames to test.
         if self.visit.coordinateSystem != FannedOutVisit.CoordSys.ICRS:
-            raise ValueError("Only ICRS coordinates are supported in Visit, "
-                             f"got {self.visit.coordinateSystem!r} instead.")
+            _log.warning("Got non-ICRS coordinates %s in Visit %s.",
+                         repr(self.visit.coordinateSystem), self.visit)
```
kfindeisen, line 329:
```suggestion
            _log.warning("Got non-ICRS coordinates %r in Visit %s.",
                         self.visit.coordinateSystem, self.visit)
```

kfindeisen: Won't this still break when trying to preload spatial datasets (refcats and templates)?

kfindeisen (changes requested):
No objections to the existing changes, but I think `prep_butler` still assumes the visit contains sky coordinates...

python/activator/middleware_interface.py
hsinfang: The nextVisit of those calibration frames have [ra,dec]=[0,0], so they don't "break" but just waste time copying non-useful data.  But it's true that we may not rely on them always being [0,0] as the default; we might tell summit to do so assuming we can have our own special "prompt-processing-test" observing script.

kfindeisen: I'm confused. What is the resolution of this issue?

hsinfang: (Sorry. I did not remember hitting "Resolve conversation" here and must have been a mistake)

hsinfang: Coming back to this after a long time... 

A bad value for `dec` indeed would break `prep_butler`, but any numeric values for `ra` and `cameraAngle` would be fine; any processed results after ISR would be suspicious of course.  I changed it so it still raises if values are obviously bad, otherwise just logs a message and continue.  

For the default case of [ra,dec]=[0,0], I think wasting time downloading potentially useless refcats is not a big deal. 

What do you think?

kfindeisen: I'd prefer to make loading of spatial datasets be conditional on having spatial coordinates in the first place, just because it's less error-prone than something that just happens to work. But I agree that run time is not a problem.

hsinfang: I hesitated on making these visits handled any differently and found the happenstance convenient, but you have a good point. It's indeed more robust to not use those coordinates at all.  The last commit has been rewritten, and now the pointless loading will be skipped for non-on-sky visits.

python/activator/middleware_interface.py
```diff
@@ -327,12 +322,22 @@ def __init__(self, central_butler: Butler, image_bucket: str, visit: FannedOutVi
...
+        # Continue the processing but using the coordinates to preload spatial datasets
+        # will be skipped. This allows cases where only ISR is wanted.
+        self._skip_spatial_preload = False
         if self.visit.coordinateSystem != FannedOutVisit.CoordSys.ICRS:
-            raise ValueError("Only ICRS coordinates are supported in Visit, "
-                             f"got {self.visit.coordinateSystem!r} instead.")
+            self._skip_spatial_preload = True
+            _log.error("Only ICRS coordinates are fully supported. "
+                       f"Got {self.visit.coordinateSystem!r} instead in {self.visit}. "
+                       "Spatial datasets won't be loaded.")
```
kfindeisen, line 335: Should this be a warning, given that we _are_ plowing ahead anyway?

python/activator/middleware_interface.py
```diff
@@ -313,6 +307,7 @@ def _collection_template(self):
     # Class invariants:
     # self.image_host is a valid URI with non-empty path and no query or fragment.
     # self._download_store is None if and only if self.image_host is a local URI.
+    # self._skip_spatial_preload indicates that the coordinates cannot be trusted.
```
kfindeisen, line 310: This is not actually an invariant...

kfindeisen (approved):
Thanks, no further objections.

python/activator/middleware_interface.py
hsinfang: I slightly prefer this to be an error for now, so (1) it does not drown in so many warnings from the pipelines, and (2) in a way it is an error as the system isn't really designed to process those visits and something may go wrong later.  I feel that this may be similar to the level of the empty qgraph when there are no templates but we continue with the next best thing.

python/activator/middleware_interface.py
hsinfang: Thanks for catching this. Wrong place to document...  😓

hsinfang: Moved away and rephrased the comments.

PR #182, opened 2024-07-15
DM-40434: Prompt Processing continues processing after timeout or termination

kfindeisen:
This PR adds configurable timeouts and grace periods to the service container, as well as a handler for TERM signals sent by Knative.

Closes #91.


dspeck1 (approved):
LGTM

PR #183, opened 2024-07-16
DM-45275: Failure to build prompt-service with new prompt-base 

hsinfang:


base/Dockerfile
```diff
@@ -1,10 +1,12 @@
 ARG STACK_TAG=d_latest
 FROM lsstsqre/centos:${STACK_TAG}
 ENV PYTHONUNBUFFERED True
+ARG kafka=python-confluent-kafka
 RUN source /opt/lsst/software/stack/loadLSST.bash \
```
ktlim, line 5: You might want to look into the relatively new ["here docs" feature](https://docs.docker.com/reference/dockerfile/#here-documents) of Dockerfiles to improve the readability of this RUN command.  (Make sure to include `set -e` to get the effect of the `&&` command separators.)

base/Dockerfile
```diff
@@ -1,10 +1,12 @@
 ARG STACK_TAG=d_latest
 FROM lsstsqre/centos:${STACK_TAG}
 ENV PYTHONUNBUFFERED True
+ARG kafka=python-confluent-kafka
```
ktlim, line 4: I think it's a little unusual to use lowercase for an argument name, but it's not wrong.

ktlim (approved):
The logic looks correct.

base/Dockerfile
kfindeisen: Do you really need `set -e`? The fourth example shows the command chaining being done on the main line, which seems cleaner.

base/Dockerfile
hsinfang: Changed it to upper case.

base/Dockerfile
hsinfang: Thanks for the reference.  This is changed to use here documents.

PR #184, opened 2024-07-23
DM-45024: Reconcile PP workflows with "quick-stack" builds

kfindeisen:
This PR fixes some old issues with the `build-base.yml` workflow and adds support for pipeline containers other than `lsstsqre/centos`.


erinleighh (approved):
Everything looks good!

PR #188, opened 2024-08-09
DM-31833: Add new loadDiaCatalogs pipelineTask

isullivan:


kfindeisen:
Looks good. Not formally approving it until we can run an integration test on `dev`.

kfindeisen (approved):
No problems in `dev`, both regular and `-noForced` pipelines run fine.

PR #189, opened 2024-08-15
Update playbook docs for Butler management

kfindeisen:
This PR updates our policy for Butler versioning and documents how to set registry permissions for the PP service accounts.

### DIFF
--- a/doc/playbook.rst
+++ b/doc/playbook.rst
@@ -231,11 +231,17 @@ The central repo for development use is located at ``s3://rubin-pp-dev-users/cen
 You need developer credentials to access it, as described under `Buckets`_.
 To run ``butler`` commands, which access the registry, you also need to set ``PGUSER=pp``.
 
+Butler Dimensions Schema Versions
+---------------------------------
+
+In general, Prompt Processing can support a range of schema versions: the lower limit is set by assumptions in Prompt Processing code, while the upper limit is set by the underlying Science Pipelines version.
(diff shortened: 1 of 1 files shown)


hsinfang (approved):

hsinfang:
Looks good to me. Thank you for the updates!

doc/playbook.rst
```diff
@@ -257,6 +263,37 @@ In our case, we want to migrate to the versions that ``/repo/embargo`` is using,
    However, when using ``butler migrate`` to update ``dimensions-config``, you should delete all existing pods to ensure that their replacements have the correct version.
    This can be done using ``kubectl delete pod`` or from Argo CD (see `Development Service`_).
 
+Updating Table Permissions
+--------------------------
+
+Some ``dimensions-config`` migrations add new tables to the Butler registry schema.
+When this happens, our service accounts need to be explicitly given permission to work with those new tables.
+
+To update permissions, use ``psql`` to log in to the registry database as the owner (``pp`` for our dev repo).
```
erinleighh, line 272: Everything looks good to me as someone who is familiar with accessing `psql`. Can we safely assume everyone using this guide will have that core knowledge? If yes, then this is good to go. If not, I think a little more instruction would be helpful.

erinleighh (approved):

doc/playbook.rst
kfindeisen: I guess it should be more self-contained. Would a link to the `Databases` section be enough? I'm not sure exactly what you mean by "accessing" `psql`, given that it's a program.

erinleighh: Yes, a link back to the Databases section would be perfect!

PR #195, opened 2024-09-11
Update Playbook docs for Butler registry permissions

kfindeisen:
This PR changes the expected registry permissions to be more minimalist; we only need delete permission for a handful of tables, and usage permission for a single sequence. Requesting only the permissions that we truly need should minimize friction with the DBAs as we move into Simonyi commissioning.

### DIFF
--- a/doc/playbook.rst
+++ b/doc/playbook.rst
@@ -290,7 +290,7 @@ If any tables are missing permissions, run:
 
 .. code-block:: psql
 
-   GRANT insert,select,update ON TABLE "<table1>", "<table2>" TO hsc_prompt,latiss_prompt,lsstcomcamsim_prompt;
+   GRANT insert, select, update ON TABLE "<table1>", "<table2>" TO hsc_prompt, latiss_prompt, lsstcomcamsim_prompt;
 
 See the `GRANT command`_ for other options.
 
--- a/doc/playbook.rst
+++ b/doc/playbook.rst
@@ -280,14 +280,17 @@ To inspect table permissions:
    set search_path to <namespace>;
    \dp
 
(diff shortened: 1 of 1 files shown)


doc/playbook.rst
```diff
@@ -280,14 +280,17 @@ To inspect table permissions:
...
 
-Most tables should grant the INSERT (a), SELECT (r), UPDATE (w), and DELETE (d) `PostgreSQL privileges`_ to all service users (currently ``latiss_prompt``, ``hsc_prompt``, and ``lsstcomcamsim_prompt``).
-Sequences should have SELECT (r) and USAGE (U) instead.
+All tables should grant the INSERT (a), SELECT (r), and UPDATE (w) `PostgreSQL privileges`_ to all service users (currently ``latiss_prompt``, ``hsc_prompt``, and ``lsstcomcamsim_prompt``).
+Some tables also need DELETE (d), but we are still confirming which ones.
+
+We need SELECT (r) and USAGE (U) permissions for the sequence ``collection_seq_collection_id``, but *not* for ``dataset_calibs_*_seq_id``, ``dataset_type_seq_id``, or ``dimension_graph_key_seq_id``.
+We expect that most future sequences will only be touched by repository maintenance and not by pipeline runs or data transfers.
 
 If any tables are missing permissions, run:
```
hsinfang, line 289: On Slack @ktlim  mentioned that these were manually applied
`GRANT SELECT,INSERT,UPDATE ON ALL TABLES IN SCHEMA embargo, embargo_old TO latiss_prompt, lsstcomcamsim_prompt;`

Should we just advise to grant on all tables, not selected tables?

hsinfang (approved):
Thanks for tracking this down and documenting it!

doc/playbook.rst
kfindeisen: I'd rather keep some human judgment in the mix -- I've seen too many people interpret "run this"-type instructions as something to obey without question. That's why I said to run `\dp` and see what's there.

PR #199, opened 2024-09-16
DM-46312: Migrate unit tests to ComCamSim data

kfindeisen:
This PR migrates the unit test repo from DECam to ComCamSim, and cleans up some related technical debt.


hsinfang (approved):
LGTM -- thanks for updating the test repo. 

`tests/data/make_central_repo.sh` seems missed in this update. I've found that file valuable in re-constructing or modifying the test repo, which we do need to do from time to time.

kfindeisen:
Oops, so that's where `export.yaml` was used! I'll fix it.

hsinfang:
Another comment: with this update, we no longer need `setupRequired(obs_decam)` in `ups/prompt_processing.table`, I think.

kfindeisen:
@hsinfang would you like to take a look at `make_central_repo.sh`? I ended up almost completely rewriting it.

hsinfang:
Looks great to me, and it will also serve as a good reference if later we want to build a new repo with another dataset. Thank you!

PR #200, opened 2024-09-26
DM-46520: Raise on AnnotatedPartialOutputsError in Prompt Processing

kfindeisen:
This PR adds behavior flags to `SingleQuantumExecutor` to be less tolerant of pipeline failures.


isullivan (approved):
Agreed that this is the behavior we want in Prompt Processing.

PR #202, opened 2024-09-27
DM-46530: Write up and test backport procedure for Prompt Processing

kfindeisen:
This PR documents how to backport bug fixes onto a release branch, and modifies the existing docs and workflows to account for the existence of release branches.


doc/playbook.rst
```diff
@@ -148,6 +149,46 @@ At the HEAD of the ``main`` branch, create and push a tag with the semantic vers
...
+.. code-block:: sh
+
+   git checkout -b releases/X.Y X.Y.0
+   git push -u origin releases/X.Y
+
+If you have a branch for your bug fix, you can make a copy for the release branch:
+
+.. code-block:: sh
+
+   git checkout -b tickets/DM-XXXXX-X.Y tickets/DM-XXXXX
```
isullivan, line 170: Here and later, did you mean:
`git checkout -b tickets/DM-XXXXX-X.Y.Z tickets/DM-XXXXX`

isullivan (commented):
The changes look good to me, but not marking as `approved` since there are still other reviewers.

doc/playbook.rst
kfindeisen: No, that was intentional -- it's the copy of `tickets/DM-XXXXX` being merged into the `X.Y` branch (by analogy to the `-v23` suffix in the Science Pipelines policy).

kfindeisen: Also, GitHub tracks each reviewer's approval/disapproval separately. 🙂

doc/playbook.rst
```diff
@@ -148,6 +149,46 @@ At the HEAD of the ``main`` branch, create and push a tag with the semantic vers
    git tag -s X.Y.Z -m "X.Y.Z"
    git push --tags
 
+Patch Releases and Release Branches
+-----------------------------------
+
+During commissioning and operations, it may be necessary to quickly deploy a bug fix without making any other changes that might potentially introduce new breakages.
+This can be done using a patch version (``X.Y.Z``).
+If there have been unrelated changes committed since the last release, you will need to isolate the bug fixes on a release branch.
+
```
hsinfang, line 158: Would it be useful to put the link https://developer.lsst.io/work/backports.html  directly in the notebook as a reference how this process is based on?

kfindeisen: I think that might be more confusing than not -- there are things there that we don't want to deal with (multiple approvals, multiple release candidates, etc.), and our build system works differently. As I mentioned on Jira, our decision flow is (might be?) a bit different too -- at least this time, we decided to backport DM-46520 as part of the process of making a release, and not as part of the process of implementing DM-46520.

(That makes me think I should switch the order of "if you have a branch..." and "cherry-pick", but I can't think of a natural wording.)

hsinfang (approved):

erinleighh (approved):

PR #210, opened 2024-10-23
DM-47067: Document config overrides in PP pipelines

kfindeisen:
This PR adds an explanation for why we only set `doProduceAlerts` in PP, and moves alert generation into the main `ap_pipe` (lsst/ap_pipe#205).


kfindeisen:
Testing cannot be done unless we merge lsst/ap_pipe#205 first...

ebellm (approved):

PR #213, opened 2024-11-01
DM-47333: Logger.exception() missing 1 required positional argument

hsinfang:


python/activator/activator.py
```diff
@@ -524,7 +524,7 @@ def process_visit(expected_visit: FannedOutVisit):
 
 
 def invalid_visit(e: InvalidVisitError) -> tuple[str, int]:
-    _log.exception()
+    _log.exception(e)
```
timj, line 527:
```suggestion
    _log.exception("Invalid visit: %s", e)
```
?

I think I'd prefer a bit of context rather than just the string text of the exception along with the stack trace.

timj (approved):

python/activator/activator.py
```diff
@@ -348,7 +348,7 @@ def next_visit_handler() -> tuple[str, int]:
         try:
             expected_visit = parse_next_visit(flask.request)
         except ValueError as msg:
-            _log.exception()
+            _log.exception(msg)
```
timj, line 351: This is existing but `msg` as an exception variable is a bit non-standard.

timj:
```suggestion
            _log.exception("Bad request: %s", msg)
```

Oh, and probably add a context message here like you have in the return.

hsinfang: Indeed. I'm changing it to `e`

python/activator/activator.py
```diff
@@ -347,9 +347,9 @@ def next_visit_handler() -> tuple[str, int]:
     try:
         try:
             expected_visit = parse_next_visit(flask.request)
-        except ValueError as msg:
-            _log.exception()
-            return f"Bad Request: {msg}", 400
+        except ValueError as e:
+            _log.exception("Bad Request: %s", e)
```
kfindeisen, line 351: Sorry, but this seems incorrect to me -- `Logger.exception` normally does not need to take the exception object, because [that's pulled automatically from the handling context](https://docs.python.org/3/library/logging.html#logging.Logger.exception). Doesn't this lead to duplicate exception messages?

hsinfang: But it does require an argument. Or do you mean removing this line here and just letting error handler handle later?

timj: No, he's saying:

```python
_log.exception("Bad request")
```
instead.

For example if you include the exception string:

```
WARNING:test:Message one
ERROR:test:Bad code: Message 2
Traceback (most recent call last):
  File "/Users/timj/work/lsstsw/build/daf_butler/y.py", line 9, in <module>
    raise RuntimeError("Message 2")
RuntimeError: Message 2
Finish
```
so you get "Message 2" in the first line of the log output and the final line of the output.

hsinfang: I see. Thank you for explaining.  
New PR: https://github.com/lsst-dm/prompt_processing/pull/214 
Would one of you review please?

PR #214, opened 2024-11-04
DM-47333: Logger.exception() missing 1 required positional argument

hsinfang:


kfindeisen (approved):
Looks good, thanks for putting up with my nit-picking! 🙂

hsinfang:
Many thanks for looking and catching the problem!

PR #215, opened 2024-11-05
DM-47387: Prompt Processing can't handle blank filters in nextVisit message

kfindeisen:
This PR adds some special-case handling for nextVisit messages with empty filters (as is common for engineering data). The cases are handled at a fairly low level (`_export_calibs` and `_export_skymap_and_templates`) because I don't see an obvious way to unify what's essentially a query tweak. This may need to be revisited on #204, which reworks how the preload queries are executed.


hsinfang (approved):
Looks good. Thanks for the idea and for the fix.
I'm worried about the growing list of dataset type names that we have to include in this code. I see there is a TODO there; maybe we want to think about it sooner.

PR #221, opened 2024-11-07
DM-46641: Test new LATISS calibs of IsrTaskLSST in PP dev butler repo

hsinfang:


hsinfang:
The config update was done in https://github.com/lsst-dm/prompt_processing/pull/220

PR #223, opened 2024-11-15
DM-47591: Allow main pipeline execution if preprocessing pipeline fails

kfindeisen:
This PR improves the error handling around pipeline execution, distinguishing PP execution failures from pipeline failures and ignoring the latter for preprocessing pipelines.

### DIFF
--- a/python/activator/exception.py
+++ b/python/activator/exception.py
@@ -22,6 +22,8 @@
 
 __all__ = ["NonRetriableError", "RetriableError", "GracefulShutdownInterrupt",
            "InvalidVisitError", "IgnorableVisit",
+           "InvalidPipelineError", "NoGoodPipelinesError",
+           "PipelinePreExecutionError", "PipelineExecutionError",
            ]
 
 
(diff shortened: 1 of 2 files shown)


python/activator/exception.py
```diff
@@ -107,3 +109,28 @@ class IgnorableVisit(ValueError):
...
     activator.visit.FannedOutVisit
     """
+
+
+class InvalidPipelineError(ValueError):
+    """Exception raised if a pipeline cannot be loaded or configured.
+    """
+
+
+class NoGoodPipelinesError(RuntimeError):
```
andy-slac, line 119: I don't see anything catching this specific exception. The `MWI._run_preprocessing` method is documented to raise this exception, and it is called from `MWI.prep_butler`, but the latter does not seem to handle `NoGoodPipelinesError` and does not document that it can raise it. Do the clients of `prep_butler` need to handle generic `RuntimeError` or `Exception`?

andy-slac (approved):
Looks good, one minor question. Devising a good exception hierarchy is hard, if anything needs to be improved in that regard in daf_butler or ctrl_mpexec, let us know.

python/activator/exception.py
kfindeisen: You are correct that `NoGoodPipelinesError` is not caught -- except where chained to another exception, it will propagate up to the top-level (Flask) exception handler. While it's an unlikely scenario in preprocessing (`prep_butler` itself controls the inputs for the pipelines we run there), I suppose we would want to follow the same policy of "try a simple main pipeline anyway".

PR #225, opened 2024-11-18
DM-47628: Prompt Processing does not fall back on missing PrerequisiteInputs

kfindeisen:
This PR fixes a bug where failure to build a quantum graph because of missing refcats is not handled by falling back to a pipeline that doesn't need them, instead causing execution to abort.

### DIFF
--- a/python/activator/middleware_interface.py
+++ b/python/activator/middleware_interface.py
@@ -1255,8 +1255,8 @@ class MiddlewareInterface:
                 with lsst.utils.timer.time_this(
                         _log, msg=f"executor.make_quantum_graph ({label})", level=logging.DEBUG):
                     qgraph = executor.make_quantum_graph(pipeline, where=data_ids)
-            except (FileNotFoundError, MissingDatasetTypeError) as e:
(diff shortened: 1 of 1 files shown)


python/activator/middleware_interface.py
```diff
@@ -1255,8 +1255,8 @@ def _try_pipelines(self, pipelines, in_collections, data_ids, *, label):
                 with lsst.utils.timer.time_this(
                         _log, msg=f"executor.make_quantum_graph ({label})", level=logging.DEBUG):
                     qgraph = executor.make_quantum_graph(pipeline, where=data_ids)
-            except MissingDatasetTypeError as e:
-                _log.error(f"Building quantum graph for {pipeline_file} failed ", exc_info=e)
+            except (FileNotFoundError, MissingDatasetTypeError):
+                _log.exception(f"Building quantum graph for {pipeline_file} failed")
```
erinleighh, line 1259:
```suggestion
                _log.exception(f"Building quantum graph for {pipeline_file} failed.")
```

erinleighh (approved):

PR #227, opened 2024-11-19
DM-47620 Add single-frame SS association to SingleFrame.yaml

Gerenjie:
My understanding of why I was instructed to make these changes in the particular structure I did:

As-is, SingleFrame.yaml exists in prompt_processing but nowhere else. We want a system in which ap_pipe (and ci_pipe and drp_pipe) sit on top of everything else and describe pipeline functionality via yamls. All prompt_processing pipeline yamls should reference and modify counterparts in ap_pipe. At the moment, SingleFrame.yaml is a standalone prompt_processing pipeline, so in addition to adding single-frame association to it, I was asked to also move it into ap_pipe in a traditional LSSTComCam -> _ingredients reference chain.

### DIFF
--- a/pipelines/LSSTComCam/SingleFrame.yaml
+++ b/pipelines/LSSTComCam/SingleFrame.yaml
@@ -1,7 +1,9 @@
-description: Single-frame pipeline specialized for LSSTComCam
+description: >-
+  Single-frame pipeline for the case in which
+  no templates exist, specialized for LSSTComCam
 
 imports:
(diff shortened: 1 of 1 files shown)


mrawls (approved):

mrawls (commented):
Suggest amending "in the event no templates exist" to description in line 1.

kfindeisen:
FTR, I asked @Gerenjie not to merge this along with the Science Pipelines changes, because we can't test until we have a daily and this change breaks compatibility with previous Stack builds. I'll try to do an _a posteriori_ test tomorrow morning.

PR #229, opened 2024-11-22
DM-47750: Disable brighter fatter in AP's ISR 

hsinfang:


pipelines/LSSTComCam/ApPipe.yaml
```diff
@@ -9,6 +9,10 @@ imports:
     include:
       - prompt
 tasks:
+  isr:
+    class: lsst.ip.isr.IsrTaskLSST
+    config:
+      doBrighterFatter: False
```
kfindeisen, line 15: Could you add a comment saying that this will be merged into `ap_pipe` on... I guess it's still DM-47750?

kfindeisen (approved):
Looks good, but I'd prefer explicitly saying this is a temporary change, just in case somebody looks.

pipelines/LSSTComCam/ApPipe.yaml
hsinfang: Thanks. Adding a comment is a good idea. 

Yes I think just using DM-47750 is good enough; no need for more paperwork :p

PR #231, opened 2024-11-22
DM-47772: Error when preloading for pipelines that don't use templates

kfindeisen:
This PR fixes an error message that appears when the `nextVisit` message has a position but the service is not configured with any pipelines that use templates.


hsinfang (approved):
LGTM

PR #234, opened 2024-11-27
DM-47829: Prompt Processing never retries if exposure does not generate visits

kfindeisen:
This PR fixes a bug that flags failed engineering runs as "potentially modifying" the APDB.

### DIFF
--- a/tests/test_middleware_interface.py
+++ b/tests/test_middleware_interface.py
@@ -478,10 +478,10 @@ class MiddlewareInterfaceTest(unittest.TestCase):
         self.interface.prep_butler()  # Ensure raw collections exist.
         filename = "nonexistentImage.fits"
         filepath = os.path.join(self.input_data, filename)
-        data_id, file_data = fake_file_data(filepath,
-                                            self.interface.butler.dimensions,
-                                            self.interface.instrument,
-                                            self.next_visit)
+        _, file_data = fake_file_data(filepath,
+                                      self.interface.butler.dimensions,
+                                      self.interface.instrument,
+                                      self.next_visit)
         with unittest.mock.patch.object(self.interface.rawIngestTask, "extractMetadata") as mock, \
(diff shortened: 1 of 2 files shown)


tests/test_middleware_interface.py
```diff
@@ -71,6 +71,9 @@
                                             "${PROMPT_PROCESSING_DIR}/tests/data/SingleFrame.yaml",
                                             ],
                               }])
+pipelines_minimal = PipelinesConfig([{"survey": "SURVEY",
+                                      "pipelines": ["${PROMPT_PROCESSING_DIR}/tests/data/ISR.yaml",],
```
isullivan, line 75: Really minor, but I think this is missing whitespace before the closing ]

tests/test_middleware_interface.py
```diff
@@ -701,11 +773,27 @@ def test_run_pipeline_cascading_exception(self):
...
+        with unittest.mock.patch(
+                "activator.middleware_interface.SeparablePipelineExecutor.pre_execute_qgraph"), \
+             unittest.mock.patch("activator.middleware_interface.SeparablePipelineExecutor.run_pipeline") \
+                as mock_run, \
+             unittest.mock.patch.object(self.interface, "main_pipelines", pipelines_minimal), \
+             unittest.mock.patch("lsst.dax.apdb.ApdbSql.containsVisitDetector") as mock_query:
+            mock_run.side_effect = ValueError("Error: not computable")
+            mock_query.return_value = False
+            with self.assertRaises(PipelineExecutionError):
+                self.interface.run_pipeline({42})
```
isullivan, line 795: A comment here reminding/explaining that group 42 was constructed as engineering-only data would be helpful.

isullivan (approved):
Looks good! A couple of very minor points.

tests/test_middleware_interface.py
kfindeisen: Flake8 doesn't complain either way. 🤔

PR #235, opened 2024-12-03
DM-47818: Leave records of calib datasets used in PP operations 

hsinfang:


isullivan (approved):

tests/test_middleware_interface.py
```diff
@@ -1001,6 +1003,29 @@ def _make_expanded_ref(registry, dtype, data_id, run):
...
+
+        Returns
+        -------
+        qgraph : `~lsst.pipe.base.QuantumGraph`
+            Quantum graph instance.
+        """
+        with tempfile.TemporaryDirectory() as tmpdir:
+            if n_quanta == 0:
+                butler = Butler(Butler.makeRepo(tmpdir))
+                return lsst.pipe.base.QuantumGraph({}, universe=butler.dimensions)
```
kfindeisen, line 1023: It seems a bit gratuitous to make a temporary repo just to get a QuantumGraph that doesn't use it. Can `DimensionUniverse` not be constructed directly?

timj: Yes. If you don't care about a specific universe then `universe=DimensionUniverse()` is all you need here.

hsinfang: Thank you. That's definitely better. 

Modified.

PR #237, opened 2024-12-03
DM-43082: Knative Keda co-existence

dspeck1:
Adds support for Knative and Keda co-existence of .  Which platform to run is determined by the PLATFORM environment variable.  Environment variables are conditionally loaded for Keda if that platform is set.  Keda will process one kafka message from fan out events with a manual commit.   Timing events come from the Kafka message header.  A timeout is defined for the Kafka message polling to obtain a fan out next visit message so that the message polling does not run continually when pods are scaled up.  The bucket notification consumer offset is changed from statically set to latest to an environment variable.

### DIFF
--- a/python/activator/activator.py
+++ b/python/activator/activator.py
@@ -27,16 +27,20 @@ import functools
 import json
 import logging
 import os
+import signal
+import socket
 import sys
 import time
-import signal
-import yaml
 import uuid
+import yaml
 
 import boto3
 from botocore.handlers import validate_bucket_name
 import cloudevents.http
 import confluent_kafka as kafka
+from confluent_kafka.serialization import SerializationContext, MessageField
+from confluent_kafka.schema_registry import SchemaRegistryClient
+from confluent_kafka.schema_registry.avro import AvroDeserializer
 import flask
 
 from .config import PipelinesConfig
@@ -53,7 +57,8 @@ from .raw import (
 from .repo_tracker import LocalRepoTracker
 from .visit import FannedOutVisit
 
-
+# Platform that prompt processing will run on
+platform = os.environ["PLATFORM"].lower()
 # The short name for the instrument.
 instrument_name = os.environ["RUBIN_INSTRUMENT"]
 # The skymap to use in the central repo
@@ -76,6 +81,31 @@ kafka_cluster = os.environ["KAFKA_CLUSTER"]
 kafka_group_id = str(uuid.uuid4())
 # The topic on which to listen to updates to image_bucket
 bucket_topic = os.environ.get("BUCKET_TOPIC", "rubin-prompt-processing")
+# Offset for Kafka bucket notification.
+bucket_notification_kafka_offset_reset = os.environ.get("BUCKET_NOTIFICATION_KAFKA_OFFSET_RESET", "latest")
+
+# Conditionally load keda environment variables
+if platform == "keda":
+    # Kafka Schema Registry URL for next visit fan out messages
+    fan_out_schema_registry_url = os.environ["FAN_OUT_SCHEMA_REGISTRY_URL"]
+    # Kafka cluster with next visit fanned out messages.
+    fan_out_kafka_cluster = os.environ["FAN_OUT_KAFKA_CLUSTER"]
+    # Kafka group for next visit fan out messages.
+    fan_out_kafka_group_id = os.environ["FAN_OUT_KAFKA_GROUP_ID"]
+    # Kafka topic for next visit fan out messages.
+    fan_out_kafka_topic = os.environ["FAN_OUT_KAFKA_TOPIC"]
+    # Kafka topic offset for next visit fan out messages.
+    fan_out_kafka_topic_offset = os.environ["FAN_OUT_KAFKA_TOPIC_OFFSET"]
+    # Kafka Fan Out SASL Mechansim.
+    fan_out_kafka_sasl_mechanism = os.environ["FAN_OUT_KAFKA_SASL_MECHANISM"]
+    # Kafka Fan Out Security Protocol.
+    fan_out_kafka_security_protocol = os.environ["FAN_OUT_KAFKA_SECURITY_PROTOCOL"]
+    # Kafka Fan Out Consumer Username.
+    fan_out_kafka_sasl_username = os.environ["FAN_OUT_KAFKA_SASL_USERNAME"]
+    # Kafka Fan Out Consumer Password.
+    fan_out_kafka_sasl_password = os.environ["FAN_OUT_KAFKA_SASL_PASSWORD"]
+    # Time to wait for fanned out messages before spawning new pod.
+    fanned_out_msg_listen_timeout = int(os.environ.get("FANNED_OUT_MSG_LISTEN_TIMEOUT", 300))
 
 _log = logging.getLogger("lsst." + __name__)
 _log.setLevel(logging.DEBUG)
@@ -127,7 +157,7 @@ def _get_consumer():
     return kafka.Consumer({
         "bootstrap.servers": kafka_cluster,
         "group.id": kafka_group_id,
-        "auto.offset.reset": "latest",  # default, but make explicit
+        "auto.offset.reset": bucket_notification_kafka_offset_reset,
     })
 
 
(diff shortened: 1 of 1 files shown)


python/activator/activator.py
```diff
@@ -195,6 +227,113 @@ def create_app():
...
+
+    _log.info("starting fan out consumer")
+
+    try:
+        fan_out_consumer.subscribe([fan_out_kafka_topic])
+
+        fan_out_listen_start_time = time.time()
+
+        while time.time() - fan_out_listen_start_time < fanned_out_msg_listen_timeout:
+            fan_out_consumer.subscribe([fan_out_kafka_topic])
```
kfindeisen, line 305: Is this code really supposed to re-subscribe every 5 seconds?

python/activator/activator.py
```diff
@@ -195,6 +227,113 @@ def create_app():
...
+            fan_out_message = fan_out_consumer.poll(timeout=5)
+            if fan_out_message is None:
+                continue
+            if fan_out_message.error():
+                if fan_out_message.error().code() == KafkaError._PARTITION_EOF:
+                    # End of partition event
+                    _log.warning('reached end of offset')
+                elif fan_out_message.error():
+                    raise KafkaException(fan_out_message.error())
```
kfindeisen, line 314: Is is really a good idea to abort on any Kafka error? We get lots of harmless errors while listening for raw notifications, that's what the `_filter_messages` function is for.

python/activator/activator.py
kfindeisen: I'm also a bit confused about the workflow. Something calls `keda_start`, and then... an actual nextVisit might arrive 300 seconds (by default) later? Does that mean this function is being run (and rerun) in the absence of messages?

python/activator/activator.py
```diff
@@ -195,6 +227,113 @@ def create_app():
...
+                                                                       SerializationContext(
+                                                                           fan_out_message.topic(),
+                                                                           MessageField.VALUE))
+                _log.info("Unpacked message as %r.", deserialized_fan_out_visit)
+
+                # Calculate time to load knative and receive message based on timestamp in Kafka message
+                _log.info("Message timestamp %r", fan_out_message.timestamp())
+                fan_out_kafka_msg_timestamp = fan_out_message.timestamp()
+                fan_out_to_prompt_time = int(time.time() * 1000) - fan_out_kafka_msg_timestamp[1]
+                _log.info("Seconds since fan out message delivered %r", fan_out_to_prompt_time/1000)
```
kfindeisen, line 326: I believe these are debugging logs, not progress logs, and should be `debug`.

python/activator/activator.py
```diff
@@ -269,6 +408,13 @@ def parse_next_visit(http_request):
     if not event.data:
         raise ValueError("empty CloudEvent received")
 
+    if platform == "knative":
+        # Calculate time to load knative and receive message based on time header from knative request
+        _log.info("Fan out send event at %s", event['time'])
+        fan_out_knative_msg_timestamp = float(event['time'])
+        fan_out_to_prompt_time = time.time() - fan_out_knative_msg_timestamp
+        _log.info("Seconds since fan out message delivered %r", fan_out_to_prompt_time)
```
kfindeisen, line 416: This code will crash in two circumstances:
- `event` is not actually a valid nextVisit event. This can be fixed by moving this block to after `visit` is constructed.
- Next-Visit Fan Out does not send a `time` header. Either we need to treat this as a breaking change in PP, or the code needs to be modified to handle events with no `time`. I'd prefer the latter, of course; it seems absurd to make a major release over some debugging code.

kfindeisen: Same comment as above about logging levels.

python/activator/activator.py
```diff
@@ -588,8 +734,16 @@ def server_error(e: Exception) -> tuple[str, int]:
 def main():
     # This function is only called in test environments. Container
     # deployments call `create_app()()` through Gunicorn.
-    app = create_app()
-    app.run(host="127.0.0.1", port=8080, debug=True)
+    if platform == "knative":
+        _log.info("starting knative instance")
```
kfindeisen, line 738: This line is misleading, since it's not running a Knative instance. Maybe something like "Starting standalone Flask app?" Possibly the same comment about the other line; like I said, I'm confused about the workflow.

kfindeisen: It would be helpful if you could update the comment at the start of `main` to say how production deployments call `keda_start` (assuming they do).

kfindeisen (changes requested):
I realize we're in a hurry to get this deployed, but I don't think the new code can provide a meaningful robustness comparison to the mainline build as-is:
* Possibly vulnerable to disk space leaks (but it's possible you don't need `LocalRepoTracker` at all; clarifying the workflow should resolve that question).
* Probably vulnerable to unhandled exceptions, which caused invalid state in fan-out and uncontrolled restarts in Knative.
* Vulnerable to transient errors in Kafka, which are very common (our existing deployments get 1-3 a night).
* Possibly some logic errors in the main message-listening loop, but that might be my unfamiliarity with Kafka.

I'm also not happy about the time header becoming a breaking change. Fortunately, that's an easy fix.

python/activator/activator.py
```diff
@@ -76,6 +83,31 @@
 kafka_group_id = str(uuid.uuid4())
 # The topic on which to listen to updates to image_bucket
 bucket_topic = os.environ.get("BUCKET_TOPIC", "rubin-prompt-processing")
+# Offset for Kafka bucket notification.
+bucket_notification_kafka_offset_reset = os.environ["BUCKET_NOTIFICATION_KAFKA_OFFSET_RESET"]
```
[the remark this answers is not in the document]
dspeck1: I will add.

python/activator/activator.py
```diff
@@ -52,8 +54,13 @@
...
 from .repo_tracker import LocalRepoTracker
 from .visit import FannedOutVisit
+from confluent_kafka.serialization import SerializationContext, MessageField
+from confluent_kafka.schema_registry import SchemaRegistryClient
+from confluent_kafka.schema_registry.avro import AvroDeserializer
+from confluent_kafka import KafkaException, KafkaError
 
-
+# Platform that prompt processing will run on
+platform = os.environ["PLATFORM"]
```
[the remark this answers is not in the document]
dspeck1: I will add.

python/activator/activator.py
```diff
@@ -76,6 +83,31 @@
...
+    # Kafka Fan Out SASL Mechansim.
+    fan_out_kafka_sasl_mechanism = os.environ["FAN_OUT_KAFKA_SASL_MECHANISM"]
+    # Kafka Fan Out Security Protocol.
+    fan_out_kafka_security_protocol = os.environ["FAN_OUT_KAFKA_SECURITY_PROTOCOL"]
+    # Kafka Fan Out Consumer Username.
+    fan_out_kafka_sasl_username = os.environ["FAN_OUT_KAFKA_SASL_USERNAME"]
+    # Kafka Fan Out Consumer Password.
+    fan_out_kafka_sasl_password = os.environ["FAN_OUT_KAFKA_SASL_PASSWORD"]
+    # Time to wait for fanned out messages before spawning new pod.
+    fanned_out_msg_listen_timeout = int(os.environ.get("FANNED_OUT_MSG_LISTEN_TIMEOUT", 300))
```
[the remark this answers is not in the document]
dspeck1: Yes.   With the current Kafka server there is no authentication or authorization.  For security ideally we will use authentication and authorization in Kafka going forward.  There are differing authentication mechanisms available and this defines them.

python/activator/activator.py
```diff
@@ -195,6 +227,113 @@ def create_app():
...
+
+def keda_start():
+
+    setup_usdf_logger(
+        labels={"instrument": instrument_name},
+    )
+
+    # Initialize local registry
+    registry = activator.repo_tracker.LocalRepoTracker.get()
+    registry.init_tracker()
```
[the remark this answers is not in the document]
dspeck1: We could do multiple workers in the same pod, but it is no configured currently to do so.  The pod would not boot without a local registry is why it is added.  I don't understand how the registry works so will have to review.

python/activator/activator.py
dspeck1: Cloud Events have header attributes and data.  The nextVisit event is sent in the data payload of the Cloud Event.  Time is sent in the Cloud Event header as an attribute.   I will update the code to allow for time not being sent.

python/activator/activator.py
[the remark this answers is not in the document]
kfindeisen: Got it. Yet another reason to streamline the config interface when we get the chance...

python/activator/activator.py
```diff
@@ -195,6 +227,113 @@ def create_app():
...
+
+    # Check initialization and abort early
+    _get_consumer()
+    _get_storage_client()
+    _get_central_butler()
+    _get_local_repo()
```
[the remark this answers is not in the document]
dspeck1: Agreed.  I will add the exception handling.

python/activator/activator.py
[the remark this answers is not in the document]
kfindeisen: Oops, my mistake. `_get_local_repo` _does_ register the local repo, so you're half using it. 🙂

The repo is unregistered by calling [`LocalRepoTracker.pop`](https://github.com/lsst-dm/prompt_processing/blob/main/config/gunicorn.conf.py#L30-L35). The trick is finding a way to guarantee it when the worker was suddenly killed, I don't know what that looks like in Keda.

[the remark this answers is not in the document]
kfindeisen: If you're not planning to configure multiple workers for now, maybe just add [a TODO comment](https://developer.lsst.io/python/style.html#to-do-comments-should-include-a-jira-issue-key) to the finally block? The issue to do it properly can be assigned to me.

python/activator/activator.py
dspeck1: Thanks for catching.  A redundant subscribe.  I will remove and sorry if that lead to the confusion. 

Keda is watching the consumer group lag on the fanned out event Kafka topic.   Keda will create new Kubernetes Jobs/Pods based on the consumer group lag and the scaling parameters (similar to the Knative scaling).   Some of the pods will be created ahead of when the next visit event arrives.  There needs to be a poll for messages to account for that.

There is a maximum run time in Keda, but the maximum run time by itself would have had Pods prematurely terminated.   For example with a 15 minute Keda timeout a Pod could be polling for 10 minutes then receives a fan out message leaving only 5 minutes to complete processing.  Providing a message polling time ensures we can set a maximum run time in Keda.   300 seconds is a guess on what to start with for the timer and will probably be adjusted down.  The Keda maximum run time is set for the message polling time + time to run processing.

kfindeisen: Just to be sure I understood that explanation: a Kubernetes job is booted and runs `keda_start`, but the job is not actually "assigned" a specific nextVisit until it reads the next one from the topic? In other words, the Keda framework is responsible only for starting/closing jobs, but not actually assigning work?

I understand re: the run time -- it's the same distinction as between Knative timeout and Gunicorn timeout, and we probably want the same solution in the long run (the config specifies the processing timeout, and the framework timeout is extrapolated from that).

dspeck1: Correct.

dspeck1:
There is not a revision concept in Keda.  ArgoCD has a [revision ](https://argo-cd.readthedocs.io/en/stable/user-guide/build-environment/) that could be exposed in an env variable.

kfindeisen:
You mean `ARGOCD_APP_REVISION` or `ARGOCD_APP_REVISION_SHORT`? I guess that might work, though it would still be harder to trace than the incrementing count we're used to. 🤔

dspeck1:
Yes.  I will on Keda's Github if they have any suggestions as well.

python/activator/activator.py
```diff
@@ -195,6 +226,114 @@ def create_app():
...
+                fan_out_to_prompt_time = int(time.time() * 1000) - fan_out_kafka_msg_timestamp[1]
+                _log.debug("Seconds since fan out message delivered %r", fan_out_to_prompt_time/1000)
+
+                # Commit message and close client
+                fan_out_consumer.commit(message=fan_out_message, asynchronous=False)
+                fan_out_consumer.close()
+
+                # Process fan out visit
+                process_visit(deserialized_fan_out_visit)
+                _log.info("Processing completed for %s", socket.gethostname())
```
kfindeisen, line 330: I recommend putting this in a finally block to be analogous to [the Knative message](https://github.com/lsst-dm/prompt_processing/blob/6f3c9f86eb420847ee8b65a7d846d6d1f1bd4957/python/activator/activator.py#L531). `process_visit` raises exceptions fairly often.

python/activator/activator.py
```diff
@@ -269,6 +408,13 @@ def parse_next_visit(http_request):
     if not event.data:
         raise ValueError("empty CloudEvent received")
 
+    if event['time']:
```
kfindeisen, line 411: Does this actually work if the field does not exist (I'm having trouble finding more than the most basic CloudEvent docs), or do you need to do something like:
```suggestion
    if 'time' in event:
```

python/activator/activator.py
```diff
@@ -586,10 +732,18 @@ def server_error(e: Exception) -> tuple[str, int]:
 
 
 def main():
-    # This function is only called in test environments. Container
-    # deployments call `create_app()()` through Gunicorn.
-    app = create_app()
-    app.run(host="127.0.0.1", port=8080, debug=True)
+    # Knative deployments call `create_app()()` through Gunicorn.
+    # Keda deployments invoke main.
```
kfindeisen, line 736: Hmm... ok. 😬 Well, maybe we can get rid of Flask runner later, I think K-T is the only one who ever used it.

kfindeisen (changes requested):
Almost good to go, but I don't think the event code is backwards-compatible. Happy to sign off once that's resolved.

python/activator/activator.py
```diff
@@ -195,6 +226,114 @@ def create_app():
...
+                # Commit message and close client
+                fan_out_consumer.commit(message=fan_out_message, asynchronous=False)
+                fan_out_consumer.close()
+
+                # Process fan out visit
+                process_visit(deserialized_fan_out_visit)
+                _log.info("Processing completed for %s", socket.gethostname())
+                break
+    finally:
+        # TODO Handle local registry unregistration
```
kfindeisen, line 333:
```suggestion
        # TODO Handle local registry unregistration on DM-47975
```

python/activator/activator.py
```diff
@@ -195,6 +225,105 @@ def create_app():
...
+        _get_consumer()
+        _get_storage_client()
+        _get_central_butler()
+        _get_local_repo()
+
+        _log.info("Worker ready to handle requests.")
+
+    except Exception as e:
+        _log.critical("Failed to start worker; aborting.")
+        _log.exception(e)
```
kfindeisen, line 264: As written, the code will continue trying to execute instead of aborting.
```suggestion
        _log.exception(e)
        sys.exit(1)
```

python/activator/activator.py
```diff
@@ -195,6 +225,105 @@ def create_app():
...
+                fan_out_consumer.close()
+
+                try:
+                    # Process fan out visit
+                    process_visit(deserialized_fan_out_visit)
+                    _log.info("Processing completed for %s", socket.gethostname())
+                    break
+                except Exception as e:
+                    _log.critical("Process visit failed; aborting.")
+                    _log.exception(e)
```
kfindeisen, line 320: An exception will cause this to loop (and likely fail over the closed consumer). Possible fix:
```suggestion
                try:
                    # Process fan out visit
                    process_visit(deserialized_fan_out_visit)
                except Exception as e:
                    _log.critical("Process visit failed; aborting.")
                    _log.exception(e)
                finally:
                    _log.info("Processing completed for %s", socket.gethostname())
                    break
```

kfindeisen (approved):
No compatibility problems, but there are two logic errors in the error handling. Please fix, and then it's ready to merge.

python/activator/activator.py
kfindeisen: From @dspeck1 on Slack:
> Removed the Knative event time and can handle with a separate ticket/PR.  I had assumed the event time header was not sent, but it is.  It is in date time format so will probably just convert it without changing fan out.

PR #243, opened 2024-12-11
DM-48087: Isr-cal pipiline is misconfigured 

hsinfang:


eigerx (approved):
Good catch!

kfindeisen:
Oops, I really should know better. Sorry about that!

hsinfang:
My bad not noticing it when reviewing

PR #244, opened 2024-12-11
DM-48095: AP pipeline does not do solar system association

kfindeisen:
This PR temporarily turns on solar system association for ComCam diaPipe, so that we can test it in production.


isullivan (approved):
This is good for tonight. As your TODO comment notes, these defaults should be moved up to `ap_pipe`. We can do more tests before making those changes.

PR #245, opened 2024-12-12
DM-48095: AP pipeline does not do solar system association

kfindeisen:
This PR removes the overrides for solar system processing, which are now centralized in lsst/ap_association#261.

This PR can only be tested once we have a base container containing the Science Pipelines changes.

### DIFF
--- a/pipelines/HSC/ApPipe.yaml
+++ b/pipelines/HSC/ApPipe.yaml
@@ -7,8 +7,3 @@ imports:
 parameters:
   # Use dataset's specific templates
   coaddName: goodSeeing
-tasks:
-  diaPipe:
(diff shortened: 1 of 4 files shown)


Gerenjie (approved):

kfindeisen:
These aren't defaults -- these are the configs used by Prompt Processing itself, as opposed to the defaults in `ap_pipe`. The HSC and ComCamSim pipelines are used for Prompt Processing development, and (as noted in the commit messages) the HSC pipeline is what we used yesterday to make sure that everything integrated correctly.

Gerenjie:
Got it. Then I agree that all instruments should run ssoAssociation by default.

PR #246, opened 2024-12-12
DM-47973: Support Keda in MiddlewareInterface._get_deployment

kfindeisen:
This PR identifies deployments by package hash by default, not just for local builds, and fixes some bugs that made the package hash less consistent than intended.

### DIFF
--- a/python/activator/middleware_interface.py
+++ b/python/activator/middleware_interface.py
@@ -387,6 +387,7 @@ class MiddlewareInterface:
         packages = lsst.utils.packages.getAllPythonDistributions()
         packagehash = hashlib.md5(usedforsecurity=False)
         for package, version in sorted(packages.items()):
+            _log_trace3.debug("Deployment includes %s %s.", package, version)
             packagehash.update(bytes(package + version, encoding="utf-8"))
 
         # APDB config is included in pipeline/task configs.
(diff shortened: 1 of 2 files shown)


python/activator/middleware_interface.py
```diff
@@ -370,26 +371,55 @@ def __init__(self, central_butler: Butler, image_bucket: str, visit: FannedOutVi
...
-            return os.environ["K_REVISION"]
-        except KeyError:
-            # If not in a container, read the active Science Pipelines install.
-            packages = lsst.utils.packages.Packages.fromSystem()  # Takes several seconds!
-            h = hashlib.md5(usedforsecurity=False)
-            for package, version in packages.items():
-                h.update(bytes(package + version, encoding="utf-8"))
-            return f"local-{h.hexdigest()}"
+        # To disambiguate all stack changes, read the active Science Pipelines install.
+        packages = lsst.utils.packages.getAllPythonDistributions()
```
timj, line 408: Do you want to add a comment that this (by design) is not including non-python conda packages? (It will include conda python packages and pipeline python packages).

kfindeisen: I didn't realize there was such a thing as a "non-Python Conda package". 😵‍💫

Honestly, I'd classify that as "known issue" rather than "by design". Given how `rubin-env` is managed (i.e., lots of updates together), it shouldn't matter in practice.

timj: One of the great things about Conda is that it gives us an easy way to install a lot of libraries. If you look at the output of `conda list` you will see lots of `lib*` packages. Also `yaml` in conda-forge is the YAML C library.

hsinfang (approved):

PR #250, opened 2025-01-23
DM-46178: Use new query system in Prompt Processing

kfindeisen:
This PR removes the workarounds added on #129 and #135 and adds new implementations of calib validity queries.


kfindeisen:
Closing as a duplicate of #249. Where are these coming from?

PR #253, opened 2025-02-10
DM-48842: Prompt Processing query errors don't report which query failed

kfindeisen:
This PR fixes some of the logs in `middleware_interface` that were left too ambiguous by the code reorganization in #249.


erinleighh (approved):
Looks good, thank you! This will be very helpful!

PR #254, opened 2025-02-11
DM-48889: Factor shared code out of Prompt Processing activator package

kfindeisen:
This PR moves general-purpose code out of `activator` (and especially the `MiddlewareInterface` class) in preparation for [DM-45988](https://rubinobs.atlassian.net/browse/DM-45988). It should not change any functionality.


bsmartradio (approved):
Everything looks good to me.

PR #260, opened 2025-02-20
DM-49019: Update Tester for Keda

dspeck1:
Add support for platform argument in upload and upload from repo testers.    Update instructions in playbook.


hsinfang (approved):
Looks good to me

PR #261, opened 2025-02-22
DM-45545: Lint fixes for rubin-env 10

timj:


.github/workflows/formatting.yaml
kfindeisen, line 1: Is this supposed to replace the existing `lint.yaml`? If not, what is the difference?

timj: Yes. It can replace it (which is why I did it) but has additional scope so that if you add ruff or black or isort configurations to your pyproject.toml it will automatically enable further checks.

timj: (Okay, I see I didn't replace it in this PR -- I intended to do it later -- you can remove lint.yaml any time).

kfindeisen: Ok, I'll add it to #257.

kfindeisen: For now, I've replaced `lint` with `check-tooling` in the repo settings.

kfindeisen: For now, I've replaced `lint` with `check-tooling` in the repo settings.

timj: check-tooling isn't doing any linting (it's finding which tooling you want it to use), you would want the lint_with_flake8 to be required if you want lint checks to pass.

PR #263, opened 2025-02-28
DM-49078: Add support for worker restarts to Keda

kfindeisen:
This PR adds periodic worker restarts to Keda. This feature is needed to mitigate memory leaks in Science Pipelines.


erinleighh (approved):
Looks good, thank you!

PR #264, opened 2025-02-28
DM-49245: Prompt Processing leaks storage on OOM (again)

kfindeisen:
This PR makes the local repo registry more robust against crashes and other unexpected states.


erinleighh (approved):
Looks good to me, thank you!

PR #266, opened 2025-03-04
DM-49293: Make Prompt Processing dataset transfers more efficient

kfindeisen:
This PR implements a suggestion to divide our Butler dataset transfers by type. It's intended as a fix for OR5, and may need to be reverted on future Middleware improvements.

### DIFF
--- a/python/activator/middleware_interface.py
+++ b/python/activator/middleware_interface.py
@@ -22,6 +22,7 @@
 __all__ = ["get_central_butler", "make_local_repo", "flush_local_repo", "make_local_cache",
            "MiddlewareInterface"]
 
+import collections
 import collections.abc
 import itertools
 import logging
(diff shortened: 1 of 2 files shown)


python/activator/middleware_interface.py
```diff
@@ -106,6 +107,39 @@ def get_central_butler(central_repo: str, instrument_class: str):
...
+    -------
+    transferred : `list` [`lsst.daf.butler.DatasetRef`]
+        The datasets transferred to ``dest_butler``.
+    """
+    refs_by_type = collections.defaultdict(set)
+    for ref in source_refs:
+        refs_by_type[ref.datasetType].add(ref)
+
+    transferred = []
+    for dtype, refs in refs_by_type.items():
```
timj, line 138: Is the idea that this means that the transaction period will be shorter because fewer files are transferred per call and so the overhead on the database locking will be reduced? 
It won't help with the file transfer speeds and will actively work against it if I enable lsst/daf_butler#1162?

kfindeisen: My understanding is that this prevents transactions from grabbing lots and lots of locks that they need, but don't need _at the same time_.

timj: It will definitely reduce the length and size of database transactions. It will, though, slow down the file transfers themselves once my ticket merges (since the point of my ticket is to do all the transfers in parallel [well with 10 workers]).

timj: I'm doing my tests on data.lsst.cloud on that ticket with 1 CPU.

timj: (and it turns out that 3 workers doing transfers is better than either 2 or 10 on Google). Suggesting the number of workers needs to be a parameter that can be tuned.

kfindeisen:
Closing, as this solution only speeds up outgoing transfers by ~25%, and incoming ones not at all. The locks appear to have been a red herring, anyway.

PR #267, opened 2025-03-04
DM-49288: Clean up temporary files when uploading.

ktlim:
Also avoid multipart uploads, which can sometimes cause problems in Ceph S3 and are generally not needed due to the number of parallel uploads we're doing anyway.

### DIFF
--- a/python/tester/upload_from_repo.py
+++ b/python/tester/upload_from_repo.py
@@ -143,7 +143,7 @@ def main():
     # exposure.
     _log.debug("Uploading with %d processes...", max_processes)
     with context.Pool(processes=max_processes, initializer=_set_s3_bucket) as pool, \
-            tempfile.TemporaryDirectory() as temp_dir:
+            tempfile.TemporaryDirectory(ignore_cleanup_errors=True) as temp_dir:
         for visit in visit_list:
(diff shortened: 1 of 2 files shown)


python/tester/upload_from_repo.py
```diff
@@ -360,7 +360,8 @@ def _upload_one_image(temp_dir, group_id, butler, ref):
                             Body=json.dumps(dict(hdul[0].header)),
                             Key=dest_key.removesuffix("fits") + "json",
                         )
-            dest_bucket.upload_file(path, dest_key)
+            with open(path, "r") as temp_file:
```
isullivan, line 363: This looks like the file is in read mode, not write.

hsinfang:
I think we want this merge to `main`, unless you have reasons not to?  Later we'll rebase `tickets/DM-48926-OR5` to pick these up.

isullivan (approved):
One comment, otherwise looks good!

python/tester/upload_from_repo.py
ktlim: Yes, the temporary file is in read mode here as it's being uploaded to the object store.

PR #270, opened 2025-03-06
DM-49327: Update MiddlewareInterface for changes in Apdb API

andy-slac:
The `containsVisitDetector` method in Apdb now needs two additional parameters to make Cassandra queries more efficient.

### DIFF
--- a/python/activator/middleware_interface.py
+++ b/python/activator/middleware_interface.py
@@ -1262,11 +1262,25 @@ class MiddlewareInterface:
         changes : `bool`
             `True` if changes have been made, `False` if retries are safe.
         """
-        data_ids = self.butler.query_data_ids(["instrument", "visit", "detector"], where=where, explain=False)
+        # Need dimension records to determine region.
+        data_ids = self.butler.query_data_ids(
+            ["instrument", "visit", "detector"], where=where, with_dimension_records=True, explain=False
+        )
         if len(data_ids) == 1:
             data_id = data_ids[0]
+            # Check if processing happened already, this needs visit timestamp.
+            # Use begin or end of timespan, whichever is defined.
+            visit_time: astropy.time.Time | None = None
+            if data_id.timespan is not None:
+                visit_time = data_id.timespan.begin or data_id.timespan.end
+            if visit_time is None:
+                # Without timespan cannot call containsVisitDetector.
+                _log.warning("No timespan defined for visit: %s. Assuming APDB modified.", data_id)
+                return True
             apdb = lsst.dax.apdb.Apdb.from_uri(self._apdb_config)
-            return apdb.containsVisitDetector(data_id["visit"], self.visit.detector)
+            return apdb.containsVisitDetector(
+                data_id["visit"], self.visit.detector, region=data_id.region, visit_time=visit_time
+            )
         elif not data_ids:
             # Engineering exposures don't produce visits, but they also can't write to the APDB.
             return False


python/activator/middleware_interface.py
```diff
@@ -1262,11 +1262,20 @@ def _check_permanent_changes(self, where: str) -> bool:
...
+        # Need dimension records to determine region.
+        data_ids = self.butler.query_data_ids(
+            ["instrument", "visit", "detector"], where=where, with_dimension_records=True, explain=False
+        )
         if len(data_ids) == 1:
             data_id = data_ids[0]
+            region = data_id.region
+            # Check if processing happened already, limit it by one month.
+            now = astropy.time.Time.now()
+            month = astropy.time.TimeDelta(30, format="jd")
```
kfindeisen, line 1274: I think this can be made much smaller, e.g. 1 hour. The point here is to see whether the run that _just failed_ touched the APDB in a way that would be hard to automatically recover.

kfindeisen (approved):
One comment re: strategy (I assume the smaller the region+time, the faster the query?)

python/activator/middleware_interface.py
andy-slac: I'll have to change this to use visit time instead of processing time, it won't need an explicit window then.

python/activator/middleware_interface.py
```diff
@@ -1262,11 +1262,25 @@ def _check_permanent_changes(self, where: str) -> bool:
...
             data_id = data_ids[0]
+            # Check if processing happened already, this needs visit timestamp.
+            # Use begin or end of timespan, whichever is defined.
+            visit_time: astropy.time.Time | None = None
+            if data_id.timespan is not None:
+                visit_time = data_id.timespan.begin or data_id.timespan.end
+            if visit_time is None:
+                # Without timespan pipelines will not be able to run, so just
+                # stop here.
+                raise ValueError(f"No timespan defined for this visit: {data_id}")
```
kfindeisen, line 1279: This comment is incorrect; this method has nothing to do with running pipelines. I suggest simply deleting it.

In any case, it's problematic to raise an exception from error handling code. The best response (same as in the case where _multiple_ data IDs are found) is to log a warning and return `True`.

kfindeisen (approved):
Please change the handling of the case where the visit timestamp is missing -- `_check_permanent_changes` is an error-handling method, not a preparatory one. Otherwise looks good.

python/activator/middleware_interface.py
andy-slac: OK, will change to warning and return True.

andy-slac:
> _check_permanent_changes is an error-handling method, not a preparatory one.

I did not actually realized that `containsVisitDetector` is only called on failure. I guess the intent here is to catch the case when database inserts fail due to existing visit/detector (e.g. due to unique constraint violation) and not to retry pipeline in that case? The issue is that unique constraints only work for SQL backend, Cassandra will overwrite existing data without raising an exception. This may be an issue for data in Cassandra (you may get a duplicate set of sources if their ids don't match), and it may also be an issue when we replicate data to SQL-based PPDB. I think safer way would be to check for an existing visit/detector before re-running pipeline, not sure if you could have a logic like that.

andy-slac:
It may be more nuanced and backend-dependent. If a rerun happens for the same visit/detector, the outcome will be different for SQL and Cassandra. The reason is that `DiaForcedSource` table primary key is `(diaObjectId, visit, detector)`. Re-run will very likely cause an insert failure for SQL backend because of identical PK. Casandra behavior is different, there are no differences between insert and update, so re-run will succeed for Cassandra. But it may cause trouble during replication of Cassandra updates to SQL PPDB, depending on timing.
I want to understand more about PP assumptions, can we talk sometime during JTM?

kfindeisen:
We discussed in person with @isullivan; part of the problem is a scenario where the fan-out system glitches and the same message gets processed by two different pods (possibly in parallel). Neither the current retry guard, nor moving it to before main pipeline execution, would guard against that case. We're considering alternatives.

PR #271, opened 2025-03-06
DM-49291: Error with non-on-sky visits

kfindeisen:
This PR fixes two bugs in how we preload data for visits that don't have a position, and reorganizes the code to make similar bugs less likely in the future.


hsinfang (approved):
Looks good and thanks for adding the unit test!

PR #274, opened 2025-04-01
DM-49427: Filter out old fanned-out messages in Keda

kfindeisen:
This PR refactors the Redis connection management to something with lower algorithmic complexity, then adds a check to both Knative and Keda that rejects messages sent too long ago.


dspeck1 (approved):
LGTM.  Nice work!

PR #275, opened 2025-04-01
DM-49822: Remove Butler from upload_images.

ktlim:


python/tester/upload_from_repo.py
```diff
@@ -216,7 +218,7 @@ def prepare_one_visit(kafka_url, group_id, butler, instrument, visit_id):
     refs : iterable of `lsst.daf.butler.DatasetRef`
         The datasets for which the events are sent.
     """
-    refs = butler.registry.queryDatasets(
+    refs = butler.query_datasets(
         datasetType="raw",
         collections=f"{instrument}/raw/all",
         dataId={"visit": visit_id, "instrument": instrument},
```
eigerx, line 224: After getting `TypeError: Butler.query_datasets() missing 1 required positional argument: 'dataset_type'` and then `KeyError: 'dataId'`,  I think this wants to be 
```
refs = butler.query_datasets(
        dataset_type="raw",
        collections=f"{instrument}/raw/all",
        data_id={"visit": visit_id, "instrument": instrument},
    )
```
However, when I make those changes I get 
```
Traceback (most recent call last):
  File "/sdf/data/rubin/user/eiger/prompt_processing/python/tester/upload_from_repo.py", line 358, in <module>
    main()
  File "/sdf/data/rubin/user/eiger/prompt_processing/python/tester/upload_from_repo.py", line 151, in main
    refs = prepare_one_visit(kafka_url, group, butler, instrument, visit)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/sdf/data/rubin/user/eiger/prompt_processing/python/tester/upload_from_repo.py", line 229, in prepare_one_visit
    for data_id in refs.dataIds.limit(1).expanded():
                   ^^^^^^^^^^^^
AttributeError: 'list' object has no attribute 'dataIds'
```

Happy to debug a little more later but I think I should switch gears to figuring out the timeout issue. Thanks for the help!!

ktlim: Oops, sorry, didn't fully take into account the new interface (and shouldn't have been doing that on this ticket anyway).

python/tester/upload_from_repo.py
```diff
@@ -148,12 +149,13 @@ def main():
         for visit in visit_list:
             group = increment_group(instrument, group, 1)
             refs = prepare_one_visit(kafka_url, group, butler, instrument, visit)
+            uris = [butler.getURI(ref) for ref in refs]
```
timj, line 152: In theory `butler.get_many_uris` is faster than calling getURI in a loop (for direct butler it will do a single database query).

ktlim: I wanted to use something like that, but I didn't see that in (daily) pipelines.lsst.io.  Did something break the docs?

timj: https://pipelines.lsst.io/v/weekly/py-api/lsst.daf.butler.LimitedButler.html#lsst.daf.butler.LimitedButler.get_many_uris

ktlim: I see, it's in the base class.  A shame it doesn't get propagated to the derived.

eigerx:
Since this ticket is better practice and works, are we ok to merge to main?

python/tester/upload_from_repo.py
```diff
@@ -331,22 +332,8 @@ def _upload_one_image(temp_dir, group_id, butler, ref):
...
-            _log.error(
-                f"{ref} has multiple artifacts and cannot be handled by current implementation"
-            )
-            for transfer in transferred:
-                os.remove(transfer.path)
-            return
-
-        path = transferred[0].path
+        path = os.path.join(temp_dir, uri.basename())
+        ResourcePath(path).transfer_from(uri, transfer="copy")
```
timj, line 336: I'm not entirely sure what this code is trying to do, but it looks like it's downloading a file from remote to posix path, optionally extracting a sidecar json, and then uploading the raw it to a new location. Is that correct?

Why isn't this doing a direct transfer from one location to another using ResourcePath? Secondly, why isn't the JSON header extracted in the remote using astropy.io.fits with ResourcePath.to_fsspec? Would simplify the code a lot.

timj: Why are we modifying the headers of raw files?

hsinfang: Because we are giving them new group ID, new exposure ID, etc to fake new exposures.  The script sends a next_visit event with that metadata, and the prompt service will attempt to process the matching exposure (via the headers).

hsinfang (approved):

PR #276, opened 2025-04-02
DM-39500: Add instrument metadata command to docs

erinleighh:


doc/playbook.rst
```diff
@@ -619,6 +619,14 @@ From ``rubin-devl``, new APDB schemas can be created in the usual way:
...
        "postgresql://rubin@usdf-prompt-processing-dev.slac.stanford.edu/lsst-devl" apdb_config_lsstcomcamsim.yaml
 
+Note that for each APDB schema created, you must also set the instrument metadata:
+
+.. code-block:: sh
+
+   apdb-cli metadata set apdb_config_latiss.yaml instrument LATISS
+   apdb-cli metadata set apdb_config_hsc.yaml instrument HSC
+   apdb-cli metadata set apdb_config_lsstcomcamsim.yaml instrument LSSTComCamSim
+
```
kfindeisen, line 629: Would it be better to package this with the `create-sql` as one example, like you did in the other repos? I can see the separate block getting overlooked.

(We might take the opportunity to trim down the example to one instrument, since with 6-7 active the reader will need to generalize anyway.)

kfindeisen (approved):
Looks good!

doc/playbook.rst
parejkoj: Are these examples, or the explicit commands that the person configuring the system will use?

kfindeisen: They're examples of explicit commands? I don't think I understand the question.

parejkoj: I had thought that the commands in the playbook were supposed to be the exact commands that the person configuring an APDB would run on USDF? As opposed to examples that they would modify to purpose.

Speaking of which, we want the LSSTCam versions in here now, don't we?

kfindeisen: I don't think it's possible or safe to run a command "on faith", without considering the circumstances.

LSSTCam-dev is a work in progress, given the difficulties in sourcing appropriate input datasets. 😛

PR #278, opened 2025-04-05
DM-49425: Mock imSim's exposure time to be replayed time  in upload test

hsinfang:


python/tester/upload.py
```diff
@@ -390,6 +401,8 @@ def upload_from_pool(visit, snap_id):
             src_blob = snap_dict[snap_id][visit]
             exposure_num, headers = \
                 make_exposure_id(visit.instrument, visit.groupId, snap_id)
+            if instrument == "LSSTCam-imSim":
+                headers.update(make_imsim_time_headers(EXPOSURE_INTERVAL))
```
isullivan, line 405: I see that `make_imsim_time_headers` also calculates time.now(), which will be slightly later than `now` calculated a few lines up from here. Is that intentional, or should you pass in `now` to `make_imsim_time_headers`?

python/tester/upload_from_repo.py
```diff
@@ -309,6 +315,8 @@ def _upload_one_image(temp_dir, group_id, butler, ref):
     instrument = ref.dataId["instrument"]
     with time_this(log=_log, msg="Single-image processing", prefix=None):
         exposure_num, headers = make_exposure_id(instrument, group_id, 0)
+        if instrument == "LSSTCam-imSim":
+            headers.update(make_imsim_time_headers(EXPOSURE_INTERVAL))
```
isullivan, line 319: Same concern about passing in `now` as for the similar line in `upload_from_raws`

isullivan (approved):
My only concern is about the current time being calculated in multiple places, which means those times will be out of sync. I am not sure if that is a problem, though.

hsinfang:
Indeed there are two, different `now` timestamps: one for the simulated next_visit events and one for the raw file headers.  The difference can't be a "problem" because that is how the reality is -- the summit send a next_visit event, but it doesn't guarantee that the actual exposure will occur at a specific time.  I'm inclined to keep them separate for now, but I should add more comments to clarify. 

Two other somewhat relevant factors: 
- The prompt service doesn't really use the `startTime` of the nextVisit yet. Maybe when we do, we want to modify our upload tests to help test the relevant features. 
- A difference between our current upload script versus the summit's plan is that summit would send the next_visit 2 exposures ahead of the time, but our script just sends next_visit and then has the exposure one at a time. Probably not the highest priority to fix but just something to keep in mind for now I think.

hsinfang:
As discussed on the jira ticket, my previous implementation was incorrect because the detectors from one same exposure ended up with different exposure time.  So I aligned them after all.  The headers now have the same exposure time as in nextVisit's `startTime`.

PR #279, opened 2025-04-07
Add status badge for Release CI workflow

kfindeisen:
This PR adds a badge to the readme, and fixes the Release CI workflow's trigger conditions to ignore merges to `main` (a job condition prevents these from actually running, but they still count as execution history).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -2,6 +2,10 @@
 prompt_processing
 #################
 
+.. image:: https://github.com/lsst-dm/prompt_processing/actions/workflows/ci-release.yaml/badge.svg?event=push
+   :target: https://github.com/lsst-dm/prompt_processing/actions/workflows/ci-release.yaml
+   :alt: Release CI Build Status
+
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -2,6 +2,11 @@
 prompt_processing
 #################
 
+.. image:: https://github.com/lsst-dm/prompt_processing/actions/workflows/ci-release.yaml/badge.svg?event=push
+   :target: https://github.com/lsst-dm/prompt_processing/actions/workflows/ci-release.yaml
+
+.. Can't think of any useful alt text
```
erinleighh, line 8: It's unfortunate that the alt text can't be auto generated based on what the badge says.

Perhaps this for alt text: Release CI Build Status Badge

erinleighh (approved):
Thanks for adding this!

README.rst
kfindeisen: I was thinking that that's exactly the sort of thing that isn't helpful in e.g. a screen reader. 😕 Maybe just "Release CI Build Status", which is at least a good name for the hyperlink?

PR #280, opened 2025-04-07
DM-49477: GracefulShutdownInterrupt does not shut down Keda workers

kfindeisen:
This PR adds correct handling of `GracefulShutdownInterrupt` to Keda. The code is rather clunky, but I don't see a way to simplify it given the need to support retries in the future.

### DIFF
--- a/python/activator/activator.py
+++ b/python/activator/activator.py
@@ -554,14 +554,14 @@ def keda_start():
                         # Process fan out visit
                         process_visit(expected_visit)
                         processing_result = "Success"
+                    except IgnorableVisit as e:
+                        _log.info("Skipping visit: %s", e)
+                        processing_result = "Ignore"
                     except GracefulShutdownInterrupt:
                         _log.error(
                             "Service interrupted.Shutting down *without* syncing to the central repo.")
                         processing_result = "Interrupted"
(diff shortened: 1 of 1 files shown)


python/activator/activator.py
```diff
@@ -554,14 +554,31 @@ def keda_start():
...
-                        sys.exit(1)
                     except IgnorableVisit as e:
                         _log.info("Skipping visit: %s", e)
                         processing_result = "Ignore"
+                    except GracefulShutdownInterrupt:
+                        # Safety net to minimize chance of interrupt propagating out of the worker.
+                        _log.error("Service interrupted.")
+                        processing_result = "Interrupted"
+                        sys.exit(1)
+                    except RetriableError as e:
```
ktlim, line 565: I guess given the TODO that these cases will diverge at some point?  Otherwise, `except RetriableError, NonRetriableError as e:` would save some duplication.

ktlim (approved):

python/activator/activator.py
kfindeisen: Yes, that's why they're different exceptions in the first place. An interrupt's recoverable _a priori_, but if it happens after APDB updates...

PR #284, opened 2025-04-08
DM-49905: HSC upload_from_repo.py error "Object of type _HeaderCommentaryCards is not JSON serializable" 

hsinfang:


eigerx (approved):
Works great, thank you!

PR #286, opened 2025-04-09
Configure to use unresolved refcats (see DM-49289)

eigerx:


pipelines/LSSTCam-imSim/ApPipe.yaml
```diff
@@ -14,3 +14,4 @@ tasks:
     config:
       # Alert distribution only runnable in PP environment.
       alertPackager.doProduceAlerts: True
+      config.astrometry.referenceSelector.doUnresolved: False
```
hsinfang, line 17: Is the leading `config.` necessary?

isullivan: Agree that the leading `config.` should not be there, but more importantly this is a config for `calibrateImage` not `diaPipe`. So to set this you would need to exclude `calibrateImage` from `ApPipe.yaml` above and redefine it here, along with all of it's custom connections and configs.

isullivan (changes requested):
Major change: the config is for `calibrateImage` not `diaPipe`

kfindeisen:
Is this still relevant?

kfindeisen:
Closing as out-of-date, since we're no longer using imSim.

PR #287, opened 2025-04-09
DM-49974: Merge PP service and init-output builds into a single workflow

erinleighh:


.github/workflows/build-service.yml
```diff
@@ -7,10 +7,12 @@ on:
       - 'releases/**'
     paths:
       # Build workflow changes
+      - '.github/workflows/build-init-job.yaml'
```
kfindeisen, line 10: This file no longer exists.

kfindeisen (approved):
Looks good, just one typo.

PR #292, opened 2025-04-15
DM-48670: Enable PP upload.py for LSSTCam

hsinfang:


eigerx (approved):
Looks good, pending successful test after fanout-dev works again.

PR #293, opened 2025-04-16
DM-48647: Change docker repo to ghcr 

roceb:


kfindeisen (changes requested):
Please also update: 
- the [tag-mangling logic in `workflows/build_base.yml`](https://github.com/lsst-dm/prompt_processing/blob/main/.github/workflows/build-base.yml#L63-L68),
- the default inputs for `build_base.yml`,
- if needed, add another clause to [`actions/get-eups-tag`](https://github.com/lsst-dm/prompt_processing/blob/main/.github/actions/get-eups-tag/action.yaml#L40-L44).

kfindeisen (approved):
Looks good!

PR #297, opened 2025-04-18
DM-50328: Prompt Processing upload.py sends band instead of physical filter for LATISS

kfindeisen:
This PR fixes an issue in `upload.py` (but not `upload_from_repo.py`) where the nextVisit provided a band when it needed a filter.


eigerx (approved):
Nice, thanks!

PR #298, opened 2025-04-18
Reconfigure ISR-Cal for latest obs_lsst

kfindeisen:
This PR turns off an option that gives the following error: `ValueError: Cannot apply quadratic crosstalk correction with doBootstrap=True.`


hsinfang (approved):
LGTM for a quick fix to the issue.  I think we will want to think about this (i.e. get pipelines folks to tell us more) and possibly pick a more appropriate fix.  But good to merge so we have a release tonight.

PR #299, opened 2025-04-21
DM-50385: Rename task labels for RFC-1088

erinleighh:


kherner (commented):
The changes all look fine, but I still get a lot of test failures when trying to run scons with this branch checked out. The main problem is that it can't find a value for `apdb_config_url`.

"Oh, maybe I should also check out the latest version of ap_pipe, since it recently changed," I said to myself. I did that, and then when trying to run scons on that, with no other changes, I hit a bunch of other test errors of the form

`ValueError: Labels {'analyzePreliminarySummaryStats'} were not found in the declared pipeline`

This isn't a problem with your changes, but probably needs to get solved to be sure everything downstream works.

kherner:
For the record, the` ValueError: Labels {'analyzePreliminarySummaryStats'}` problem was due to not having the latest version of analysis_tools set up while testing. That fixed this particular issue  and then a scons run of prompt_processing fails in test_middleware_interface.py:

```
 E                   AssertionError: <DatasetExistence.UNRECOGNIZED: 0> is not true
tests/test_middleware_interface.py:351: AssertionError
```

kherner:
Looking at the latest commit, I think everything is as ready to go as it can be at this point. The templates have to be updated for the tests to pass.

kherner (approved):

PR #301, opened 2025-04-23
DM-50472: Invalid Isr-cal.yaml

hsinfang:


pipelines/LSSTCam/Isr-cal.yaml
```diff
@@ -10,6 +10,7 @@ tasks:
   isr:
     class: lsst.ip.isr.IsrTaskLSST
     config:
+      ampNoiseThreshold: .nan
```
kfindeisen, line 13: I'm surprised this works...

Is the "." needed?

kfindeisen (approved):
Looks good! Sorry again for missing this.

pipelines/LSSTCam/Isr-cal.yaml
hsinfang: Without the "." it's taken as a `str` 
```TypeError: Value nan is of incorrect type str. Expected type float```

PR #302, opened 2025-04-24
DM-50500: Prompt Processing uses incorrect definition of `SingleFrame`

kfindeisen:
This PR corrects the definition of our `SingleFrame` pipelines to be more aligned with `ap_pipe`'s.

### DIFF
--- a/pipelines/HSC/SingleFrame.yaml
+++ b/pipelines/HSC/SingleFrame.yaml
@@ -1,9 +1,12 @@
 description: >-
   Single-frame pipeline for the case in which
   no templates exist, specialized for HSC
+  Unlike ap_pipe/SingleFrame.yaml, this pipeline is not standalone.
 
 imports:
   - location: $AP_PIPE_DIR/pipelines/HSC/SingleFrame.yaml
-    include:
-      - apPipeSingleFrame
-      - analyzeCalibrationMetrics
+    exclude:
+      # Run in prep_butler or Preprocessing.yaml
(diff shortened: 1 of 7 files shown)


pipelines/HSC/SingleFrame.yaml
```diff
@@ -1,9 +1,12 @@
...
 imports:
   - location: $AP_PIPE_DIR/pipelines/HSC/SingleFrame.yaml
-    include:
-      - apPipeSingleFrame
-      - analyzeCalibrationMetrics
+    exclude:
+      # Run in prep_butler or Preprocessing.yaml
+      - getRegionTimeFromVisit
+      - mpSkyEphemerisQuery
+      - loadDiaCatalogs
```
isullivan, line 12: It is true we don't want to run `loadDiaCatalogs` in SingleFrame, but I don't actually see that in the imports so I think the exclusion is unnecessary.

isullivan (approved):
One question about excluding `loadDiaCatalogs`

pipelines/HSC/SingleFrame.yaml
kfindeisen: Yes, it's not there. This was intended as "future-proofing" against changes to `SingleFrame` (which I think is part of what caused this mess in the first place). We want to include any changes to upstream `SingleFrame`, _except_ changes that incorporate more than preprocessing.

And of course I tested that excluding a nonexistent task is legal. 🙂

PR #303, opened 2025-04-25
DM-50439: Remove hacks from DM-50385 and add template_coadd templates to unit test repo

erinleighh:


isullivan (approved):
Looks like this successfully reverts DM-50385 and adds the new template_coadds

PR #305, opened 2025-04-30
DM-50586: Include current day_obs in all Prompt Processing logs

kfindeisen:
This PR does some more factoring of the Keda event loop, then adds a `day_obs` value to all logs inside the loop. In the unlikely event that we're processing a daytime visit across the day_obs boundary (8:00/9:00 CLT/CLST), the logs will consistently use the old day.


erinleighh (approved):
Everything looks good!

PR #306, opened 2025-05-01
DM-50636: Prompt Processing unit tests need preprocessing outputs for main pipeline

kfindeisen:
This PR fixes unit tests to (mostly) self-consistently handle preprocessing and main pipelines. The bug was exposed by `w_2025_18`, and this fix is needed to make builds against that Science Pipelines version or any later ones.


hsinfang (approved):
LGTM

PR #307, opened 2025-05-07
DM-50751: Remove prompt-proto-service references from Playbook

kfindeisen:
This PR removes the remaining documentation specific to `prompt-proto-service` or Knative.


dspeck1 (approved):
LGTM

PR #312, opened 2025-05-27
DM-51080: Uploaders may duplicate group IDs

kfindeisen:
This PR fixes a bug in uploaders' group recognition code that sometimes caused group IDs to be reused, causing PP to import the wrong image.


erinleighh (approved):
Looks good!

PR #315, opened 2025-05-30
DM-47970: Make raw deletion more thorough.

erinleighh:


kfindeisen:
Is this still relevant?

kfindeisen (approved):
Please don't double-log exceptions, otherwise looks good.

kfindeisen (changes requested):

python/activator/middleware_interface.py
```diff
@@ -1851,30 +1851,30 @@ def _query_datasets_by_storage_class(self, butler, exposure_ids, collections, st
...
-                )
-                _log_trace.debug("Removing %d raws for exposures %s.", len(raws), exposure_ids)
+
+            # Clean out raws
+            raws = self.butler.query_datasets(
+                "raw",
+                collections=self.instrument.makeDefaultRawIngestRunName(),
+                find_first=False,
+                explain=False,  # Raws might not have been ingested.
+                instrument=self.visit.instrument,
```
kfindeisen: Keeping instrument in the query is probably harmless (each worker should only ever have had one!), but is it needed?

kfindeisen (approved):
Looks good (for real, this time)!

PR #317, opened 2025-06-06
DM-50617: Abort Prompt Processing if camera rotation different from nextVisit

kfindeisen:
This PR compares the sky rotation angles in the nextVisit message and the image header, and rejects the image for processing if they don't match. This prevents rotated visits from being partially processed, depending on the exact HTM/patch coverage involved.

### DIFF
--- a/python/activator/activator.py
+++ b/python/activator/activator.py
@@ -26,6 +26,7 @@ import contextlib
 import functools
 import json
 import logging
+import math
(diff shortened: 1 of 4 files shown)


erinleighh (approved):
Looks good! Comments/docs were understandable and I was able to follow the logic smoothly. Thank you!

PR #320, opened 2025-06-13
DM-51134: Migrate away from Mamba in Prompt Processing build infrastructure

kfindeisen:
This PR replaces mamba with conda and otherwise brings the base container up to date.


hsinfang (approved):
Looks good, and thanks for the detailed commit messages!

PR #322, opened 2025-06-17
DM-51389: Expand use of retry in Prompt Processing

kfindeisen:
This PR adds retries of Butler operations on `InterfaceError` and Redis operations on `ConnectionError`.

### DIFF
--- a/python/activator/middleware_interface.py
+++ b/python/activator/middleware_interface.py
@@ -84,7 +84,13 @@ padding = float(os.environ.get("PRELOAD_PADDING", 30))
 repo_retry = float(os.environ.get("REPO_RETRY_DELAY", 30))
 
 
-@connect.retry(2, sqlalchemy.exc.OperationalError, wait=repo_retry)
+# TODO: revisit which cases should be retried after DM-50934
+# TODO: catch ButlerConnectionError once it's available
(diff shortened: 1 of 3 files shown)


python/activator/middleware_interface.py
```diff
@@ -84,7 +84,11 @@
 repo_retry = float(os.environ.get("REPO_RETRY_DELAY", 30))
 
 
-@connect.retry(2, sqlalchemy.exc.OperationalError, wait=repo_retry)
+SQL_EXCEPTIONS = (sqlalchemy.exc.OperationalError, sqlalchemy.exc.InterfaceError)
```
hsinfang, line 87: Is this to be replaced by `ButlerConnectionError`?  Not sure which ticket that will be; if you know, it might be good to add a note. (DM-50934? Or does PP stop retrying these once Butler retries internally?)

hsinfang (approved):

python/activator/middleware_interface.py
kfindeisen: I honestly don't know what Middleware's plans are (as I pointed out on DM-50934, retry behavior really needs to be handled at the application level because one size doesn't fit all). But my understanding [from this conversation](https://rubin-obs.slack.com/archives/C08CANY4B6H/p1749579803107309?thread_ts=1749571090.664209&cid=C08CANY4B6H) is that `ButlerConnectionError` is meant to be caught by external clients.

PR #323, opened 2025-06-20
DM-20746: Defer to camera geometry instead of raw formatter for intiial WCS.

TallJimbo:
This is a small cleanup (reducing reliance on a middleware interface we'd prefer to stop treating as public) that should only be merged after the 'camera' datasets in all LATISS repos have been rebuilt with post-DM-20746 code.

### DIFF
--- a/python/activator/middleware_interface.py
+++ b/python/activator/middleware_interface.py
@@ -49,6 +49,7 @@ from lsst.daf.butler import Butler, CollectionType, DatasetType, DatasetRef, Tim
 import lsst.dax.apdb
 import lsst.geom
 import lsst.obs.base
+from lsst.obs.base.utils import createInitialSkyWcsFromBoresight
 import lsst.pipe.base
 from lsst.pipe.base.quantum_graph_builder import QuantumGraphBuilderError
 import lsst.analysis.tools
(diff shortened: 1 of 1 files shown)


cmsaunders:
Rereading the comments on the Jira ticket, I see that this is not supposed to be merged with the other PRs.

kfindeisen:
Is this still relevant?

(Also, typo in the commit message)

kfindeisen:
We are still using LATISS data for integration testing, but I have no idea how to recognize a "rebuilt camera".

Will it fail spectacularly if we try to run it on an incompatible repo?

kfindeisen:
Basic tests (`upload.py`) work fine. A larger run of 38 LATISS visits produced 3 instances of `NoPsfStarsToStarsMatchError`, 2 of `ObjectSizeNoGoodSourcesError`, and 1 of `BadSubtractionError`.

A run of 38 (independently selected) LATISS visits on `latest` produced 5 instances of `NoPsfStarsToStarsMatchError`, 2 of `ObjectSizeNoGoodSourcesError`, 2 of `BadSubtractionError`, and 1 of `PsfexTooFewGoodStarsError`. So I think we're good?

TallJimbo:
Yeah, let's go ahead and merge it now.  I see that the LATISS camera in `/repo/main` has still not been updated, but your testing says that shouldn't matter, and thinking about it a little more, I think I understand why (it's because LATISS has only one detector, so flipping the focal plane parity doesn't move detectors around the way it would for other cameras, and the affine initial WCS can soak up the parity flip).

PR #328, opened 2025-07-02
DM-50087: Preload IERS data in Prompt Processing

kfindeisen:
This PR implements a two-part system for caching IERS data at USDF and importing it into Prompt Processing at pod start. This is expected to be faster than a direct download from IERS/ObsPM, and ensures that we do not download in the middle of processing (though by happening at pod start it will still reduce our response time at the start of a sequence).


bin.src/export_iers.py
```diff
@@ -0,0 +1,65 @@
...
+    logging.basicConfig(level=logging.INFO, stream=sys.stdout)
+
+    args = _make_parser().parse_args()
+    # Update download cache
+    for url in IERS_URLS:
+        astropy.utils.data.download_file(url, cache="update")
+    with tempfile.NamedTemporaryFile(suffix=".zip") as local_cache:
+        astropy.utils.data.export_download_cache(local_cache, IERS_URLS, overwrite=True)
+        dest = lsst.resources.ResourcePath(args.destination)
+        src = lsst.resources.ResourcePath(local_cache.name, isTemporary=True)
```
ktlim, line 59: Is `local_cache.ospath` safer/better?

ktlim (approved):

bin.src/export_iers.py
kfindeisen: I don't think there is an `ospath` member -- it's not in [the `NamedTemporaryFile` docs](https://docs.python.org/3.11/library/tempfile.html#tempfile.NamedTemporaryFile), and when I try with our flavor of Python the only public members are the documented ones: `close`, `file`, and `name`.

PR #339, opened 2025-08-07
DM-52109: Try not to overlap exposure IDs with summit observations 

hsinfang:


python/tester/utils.py
```diff
@@ -117,7 +117,7 @@ def get_last_group(bucket, instrument, date):
         if numbers:
             last_number = max(numbers)
         else:
-            last_number = 0
+            last_number = 90000
```
kfindeisen, line 120: If this is a six-digit counter (per line 105), should this be 900_000 instead of 90_000?

kfindeisen (approved):
Looks good, just a quibble about the digit.

python/tester/utils.py
hsinfang: This is really `90_000`. 

One real life example of summit LSSTCam exposure ID=`2025072800142`, group=`2025-07-29T03:06:07.417`, obs_id=`MC_O_20250728_000142`. 

An example from our dev repo: exposure=`2025080600012`, group=`2025-08-06T00:00:00.000012`, obs_id=`MC_O_20250806_000012`. 

The 6 digits as per line 105 was just convenient as microseconds.  

And indeed `obs_id` has 6 digits there too.

hsinfang: However I missed to see if I could use a different controller in the mocked exposure.  I'll look into that more.

ktlim: The Camera image name field has 6 digits, but the exposure ids in DM have only 5 as we didn't believe there could ever be more than 100K per day (more than one per second).  So you shouldn't use sequence numbers greater than 99999.

PR #341, opened 2025-08-08
DM-52109: Update unit tests for new LSSTCam-dev conventions.

kfindeisen:
This PR updates the unit tests to pass with #339. That PR passed because of an oversight in the GitHub workflow.


hsinfang (approved):
Thanks for fixing these!

PR #347, opened 2025-09-16
DM-52461: Dev Kafdrop address

dspeck1:
Update kafdrop dev address in playbook.


kfindeisen:
Tagging this PR with `ignore-for-release`, since it's Playbook-only, and retroactively removing it from the 11.0.0 release notes.

PR #348, opened 2025-09-17
DM-51756: Refactor Prompt Processing activator module

kfindeisen:
This is a dummy PR for integration testing of Prompt Processing during the activator refactor.

### DIFF
--- a/python/activator/activator.py
+++ b/python/activator/activator.py
@@ -77,23 +77,23 @@ s3_endpoint = os.environ["S3_ENDPOINT_URL"]
 # URI of raw image microservice
 raw_microservice = os.environ.get("RAW_MICROSERVICE", "")
 # Bucket name (not URI) containing raw images
-image_bucket = os.environ["IMAGE_BUCKET"]
+raw_image_bucket = os.environ["IMAGE_BUCKET"]
 # Time to wait after expected script completion for image arrival, in seconds
 image_timeout = int(os.environ.get("IMAGE_TIMEOUT", 20))
 # Absolute path on this worker's system where local repos may be created
-local_repos = os.environ.get("LOCAL_REPOS", "/tmp")
+local_repo_space = os.environ.get("LOCAL_REPOS", "/tmp")
 # Kafka server for raw notifications
-kafka_cluster = os.environ["KAFKA_CLUSTER"]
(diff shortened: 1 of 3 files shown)


tests/test_activator.py
```diff
@@ -0,0 +1,312 @@
...
+                                    scriptSalIndex=3,
+                                    dome=FannedOutVisit.Dome.OPEN,
+                                    duration=35.0,
+                                    totalCheckpoints=1,
+                                    private_sndStamp=float(now.value) - 600.0,
+                                    )
+
+    def test_long_expiration(self):
+        self.assertTrue(is_processable(self.visit, 3600.0))
+        self.assertTrue(is_processable(self.visit, 610.0))  # Test may have long delays in Jenkins
```
hsinfang, line 239: Does this mean that we'll see whether Jenkins has a long delay and fail this test?

tests/test_activator.py
```diff
@@ -0,0 +1,312 @@
...
+
+    def test_short_expiration(self):
+        self.assertFalse(is_processable(self.visit, 100.0))
+        self.assertFalse(is_processable(self.visit, 499.9))
+
+    def test_mid_expiration(self):
+        # Right now, messages are rejected based on message age, not time since exposure.
+        # This may change in the future.
+        self.assertFalse(is_processable(self.visit, 550.0))
+        self.assertFalse(is_processable(self.visit, 510.0))  # Test may have long delays in Jenkins
```
hsinfang, line 249: I'm not sure why this line gets this comment. May you please elaborate?

hsinfang (approved):
Thanks for the well-organized commits which made the moving clear!

tests/test_activator.py
kfindeisen: No, other way around: I gave the test a huge margin on that side so that people wouldn't complain if the nightly integration tests failed. See #343 and [this Slack thread](https://rubin-obs.slack.com/archives/C07QGMAN6RL/p1755240688153089).

Would something like "Pad for possible Jenkins runtime" be clearer?

kfindeisen: Thought comparing to #343 (which added 0.1 s), maybe I made the margin a little too big?

hsinfang: Got it.  Thanks for explaining. Yes, "Pad for possible slow run time" would be clearer.

PR #379, opened 2026-05-21
DM-53590: a script to copy and link template_coadd from main to embargo

hsinfang:


hsinfang:
The new script `link_templates.py` shares similarities with the current `curate_templates.py` script in the `main` branch.  Given https://github.com/lsst-dm/prompt_processing/pull/351 is transforming/expanding the purpose of `curate_templates.py` in a different direction and they will have different purposes, to avoid conflicts I'm adding a new script `link_templates.py`  rather than reusing `curate_templates.py`. If one becomes obsolete we can remove it later or think about combining.

